PR #65, opened 2026-04-11
refactor: implement build_croissant protocol on all FileTypeHandlers

rafiattrach:
## Why

`MetadataGenerator.generate_metadata()` had grown to 647 lines with format-specific dispatch — `if "column_types" ... elif "signal_types" ... elif "image_properties"` — baked directly into the generator. Every new handler required editing the generator. This doesn't scale.

## What changed

Added `build_croissant(file_metas, file_ids) -> (FileSets, RecordSets)` as an `@abstractmethod` on `FileTypeHandler`. The generator now does exactly two things:

1. Creates all `FileObject`s (owns ID assignment and the global counter)
2. Calls `handler.build_croissant()` per handler — handlers return FileSets + RecordSets, never FileObjects

Format knowledge — column types, signal fields, image glob patterns, partitioned Parquet grouping — lives entirely inside each handler.

## Design

Three required methods per handler: `can_handle`, `extract_metadata`, `build_croissant`. The registry pattern + Strategy pattern means adding a new format is: subclass, implement three methods, register — zero other files change.

`_build_fields` (Arrow schema → `mlc.Field`, handles nested lists/structs recursively) moved to `handlers/utils.py`. Renamed `file_objects` → `distributions` in the generator to match Croissant spec terminology (`distribution` holds both FileObjects and FileSets).

## Why `open_targets_like_croissant.jsonld` changed

Previously standalone Parquet RecordSets were emitted inline during the main loop while partitioned ones were deferred post-loop. Now all Parquet output is emitted together in the post-loop batch phase. The JSON-LD is semantically identical — the Croissant spec does not require ordering, and all structural test assertions pass unchanged.

## Result

- `metadata_generator.py`: 647 → 249 lines, zero format-specific key checks
- 72/72 tests pass
- Adding a new format: subclass `FileTypeHandler`, implement 3 methods, register in `registry.py`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -176,5 +176,8 @@ cython_debug/
 # AI assistant context files
 CLAUDE.md
 
+# Planning
+.planning/
+
 # Large reference outputs not under version control
 tests/data/output/open_targets_official_croissant.json
--- a/src/croissant_baker/handlers/base_handler.py
+++ b/src/croissant_baker/handlers/base_handler.py
@@ -59,7 +59,11 @@ class FileTypeHandler(ABC):
         pass
 
     @abstractmethod
-    def build_croissant(self, file_metas: list, file_ids: list) -> tuple:
+    def build_croissant(
+        self,
+        file_metas: list[dict],
+        file_ids: list[str],
+    ) -> tuple[list, list]:  # (list[FileSet], list[RecordSet])
         """Build Croissant FileSets and RecordSets for all files this handler processed.
 
         Called once per handler after the FileObject loop. Receives all metadata
--- a/src/croissant_baker/handlers/base_handler.py
+++ b/src/croissant_baker/handlers/base_handler.py
@@ -8,12 +8,16 @@ class FileTypeHandler(ABC):
     """
     Abstract base class for file type handlers.
 
-    Each handler is responsible for:
-    - Detecting if it can process a specific file type
-    - Extracting comprehensive metadata from files it handles
+    Each handler is responsible for three things:
+    - can_handle: decide if this handler owns a given file
+    - extract_metadata: extract raw metadata from a single file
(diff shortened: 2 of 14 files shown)


rajna-fani:
Overall: Really nice work on this refactor; the direction is clear, the responsibilities make sense, and it’ll be much easier to add formats going forward. I just had a few minor revisions:

1. metadata_generator.py: meta_handler_map[id(meta)]
Using id(meta) as a dictionary key is fragile: it assumes each metadata dict keeps the same object identity for the whole generate_metadata() run. If those dicts are ever copied, rebuilt, or otherwise replaced, the handler association can silently break.
Suggestion: Avoid identity-based mapping. For example, store (handler, meta) pairs (or a small struct) in the list you build during extraction, so each row is explicitly tied to its handler without a parallel id(meta) map.

2. base_handler.py build_croissant signature:                                                                               
Bare list and tuple annotations on the public protocol method don't communicate the contract to someone implementing a new handler.  Maybe it should be list[dict], list[str], tuple[list, list] at minimum, with a comment on what the tuple contains.                               
                  
3. metadata_generator.py __init__ and save_metadata:                                                                      
Parameter docstrings were dropped from __init__ and save_metadata. These are the main API surface for external users — worth keeping the Args/Raises docs. 

4. Tests — Maybe some small, direct tests for build_croissant per handler would make the new protocol obvious and catch regressions.

rafiattrach:
Thanks! 

- `id(meta)` → store `(handler, meta)` tuples directly, no parallel identity map                               
- Type annotations → `list[dict]`, `list[str]`, `tuple[list, list]` with inline comment on contents              
- docstrings restored                                                               
- Batch dispatch wrapped in try/except to match extraction loop resilience      
- added `build_croissant` tests directly to each handler's existing test file rather than a separate one, so they live next to the handler.

rajna-fani (approved):

PR #66, opened 2026-04-12
feat: Open Targets external evaluation and committed test subset

rajna-fani:
Adds an external evaluation framework that compares Croissant Baker output against independently authored Croissant metadata, using the [Open Targets Platform](https://platform.opentargets.org) (55 Parquet datasets) as the first case.
- **`eval/` directory** — self-contained evaluation layout: download script, ground truth, evaluation script, and output artifacts. Separated from the CI test suite (`testpaths = ["tests"]`) and invoked explicitly.
- **Committed Open Targets subset** — 3 real Open Targets datasets (~2 MB) committed under `tests/data/input/open_targets_subset/` for fast CI coverage of nested-struct schemas and reserved-name subfields without any download.
- **Stabilized test outputs** — pinned `--date-published` and `--citation` in `test_open_targets_subset` and `test_reserved_field_names` so committed JSON-LD outputs are deterministic across machines and runs.
### Evaluation results (55 datasets, 2,255 Parquet files)
| Metric | Value |
|--------|-------|
| RecordSets matched | 55/55 (100%) |
| Fields matched | 819/819 (100%) |
| Strict type agreement | 693/819 (84.6%) |
| Semantic type agreement | 798/819 (97.4%) |
| True mismatches | 21/819 (2.6%) |
The 105 strict-only mismatches are precision differences (e.g., `Float` vs `Float64`) where Baker infers the exact Parquet schema type while the ground truth uses generic Croissant types. The 21 true mismatches are all `Float` (ground truth) vs `Int64` (generated) on integer-valued columns — Baker is more precise here, as the Parquet schema stores these as Int64 and the ground truth likely reflects Pandas' nullable-int-to-float coercion.
### How to run
```bash
# CI tests (no download needed)
uv run pytest tests/ -v
# Full evaluation (~20-30 GB one-time download)
bash eval/open_targets/download.sh
uv run pytest eval/open_targets/ -v

### CODE REVIEW

File: tests/test_reserved_field_names.py [hunk 1]
```diff
@@ -97,6 +97,14 @@ def test_reserved_name_struct_fields_validate(
             "Reserved-name regression test",
             "--creator",
             "Test Author",
+            "--url",
+            "https://example.com/reserved-name-test",
+            # Fixed metadata so committed tests/data/output/reserved_names_croissant.jsonld
+            # does not change on every run (defaults use datetime.now() and current year).
```
rafiattrach: nit: maybe this comment isn't needed? The pinned flags (`--date-published`, `--citation`) seem self-explanatory, and the reasoning is already in the commit history.

pyproject.toml
```diff
@@ -43,6 +43,8 @@ packages = ["src/croissant_baker"]
 
 [tool.pytest.ini_options]
 testpaths = ["tests"]
+# Ensure `src/` layout resolves when a global PYTHONPATH shadows the editable .pth
```
rafiattrach, line 46: Worth a quick note on what prompted this: was there a specific environment where the editable install wasn't resolving? Happy to keep it either way, just curious if this was a one-off or something we should expect broadly.

rafiattrach (approved):
overall looks great! Just a few minor comments

pyproject.toml
rajna-fani: It was added after pytest picked up the wrong package when PYTHONPATH shadowed the editable install (e.g. old croissant-maker or a stale env). It’s the usual src/ layout fix for pytest and is harmless when the env is clean

PR #34, opened 2026-03-25
Added RAI metadata support with PROV-O, using a YAML interface.

JoanGi:
## Add RAI metadata extension with PROV-O provenance support

This PR aims to initiate discussion on how we can add Croissant RAI metadata support to croissant-maker and, more broadly, to the set of tools in this ecosystem (such as the editor). Let me know what you think!

### Motivation

RAI metadata presents a unique challenge compared to the rest of the Croissant spec:

- Most attributes live at the dataset level and cannot be automatically extrapolated from the data structure — fields like `rai:SocialImpact` or `rai:Biases` require deliberate authorship.
- Many natural-language attributes require critical thinking by dataset authors; some of the seed works the RAI spec is inspired by reflect this human judgement.
- Newer mechanisms, such as PROV-O provenance, allow us to organise this information in a more machine-readable way — and tools should support it. Note that the last published release of `mlcroissant` (1.0.22) does not yet include PROV-O support; this feature requires installing from the latest commit.

### Approach

This PR addresses the above by proposing a **YAML interface** that lets authors create their RAI annotations in a human-friendly format, while croissant-maker handles the translation into spec-compliant, machine-actionable Croissant JSON-LD.

To faciliatate the user's journey, a teamplate is provided at the root of the project (please, change it at your desire), and an example is proposed at _tests/data/input/mimiciv_demo/physionet.org/mimiciv_demo-rai-example.yaml_

###  How to use it
When generating a new Croissant file, pass --rai-config:

```bash
croissant-maker --input ./my-dataset --creator "Name" --rai-config rai.yaml
```

When enriching an existing Croissant file, use the new rai-apply subcommand:

```bash
croissant-maker rai-apply dataset.jsonld --rai-config rai.yaml
```

 or write to a separate file:

```bash
croissant-maker rai-apply dataset.jsonld --rai-config rai.yaml --output dataset-rai.jsonld
```

### Examples and templates

- `rai-example.yaml` — fully documented template with every supported field and inline comments explaining the RAI/PROV-O mapping
- `tests/data/input/mimiciv_demo/physionet.org/mimiciv_demo-rai-example.yaml` — real-world example using the MIMIC-IV demo dataset
- `tests/data/output/mimiciv_demo_croissant_rai.jsonld` — reference output showing the resulting Croissant JSON-LD

### Changes

- `src/croissant_maker/rai/` — new module with schema (Pydantic models), YAML loader, and injector
[...]

PR #69, opened 2026-04-14
feat: add native RAI CLI flags alongside YAML workflow

rajna-fani:
This PR adds a native CLI path for Croissant RAI metadata while keeping the existing `--rai-config` YAML workflow intact.

Users now have two ways to author RAI metadata:
- direct `--rai-*` flags for the dataset-level fields already supported by `mlcroissant`
- `--rai-config` YAML for the richer provenance/activity workflow added in PR #34

## What changed

- Added native `--rai-*` flags to the main `croissant-baker` generate command
- Passed native RAI fields through to `mlcroissant.Metadata(...)`
- Added validation and normalization for text, list, and ISO datetime inputs
- Ensured RAI-bearing outputs declare `http://mlcommons.org/croissant/RAI/1.0` in `conformsTo`
- Kept the existing YAML-based `--rai-config` and `rai-apply` workflows working
- Added tests for:
  - native RAI flag generation
  - invalid timeframe parsing
  - conflict detection between native flags and YAML
  - RAI conformance for both workflows

## Notes

- Native `--rai-*` flags and `--rai-config` are intentionally mutually exclusive in this PR to keep the UX clear and avoid mixing two different RAI authoring models.
- `uv.lock` was updated to reflect the existing `pyyaml` dependency state.

## Validation

- `pytest tests/test_cli.py tests/test_rai.py`
- `ruff check src/croissant_baker/__main__.py src/croissant_baker/metadata_generator.py tests/test_cli.py`


tompollard (approved):
Looks good to me, thanks Rajna!

I notice that there are slight differences to the injector (e.g. `rai:dataBias` vs the `rai:dataBiases` used here). 

Looking at the spec (https://github.com/mlcommons/croissant/blob/main/docs/croissant-rai-spec.md), `dataBiases` appears to be correct. We should fix the injector to match.

rafiattrach (approved):

PR #70, opened 2026-04-15
rai:dataBias -> rai:dataBiases and rai:socialImpact -> rai:dataSocialImpact

tompollard:
The YAML-based RAI workflow uses `rai:dataBias` (singular) as the key for data bias. The `--rai-*` flags introduced in https://github.com/MIT-LCP/croissant-baker/pull/69 uses `rai:dataBiases` (plural).

According to the online spec at https://docs.mlcommons.org/croissant/docs/croissant-rai-spec.html, the plural `rai:dataBiases` is correct. `rai:socialImpact` also appears incorrect (according to the spec, it should be `rai:dataSocialImpact`). 

This pull request updates the YAML workflow to use `rai:dataBiases` and `rai:dataSocialImpact`. @JoanGi, could you confirm that these keys are correct?

### DIFF
--- a/rai-example.yaml
+++ b/rai-example.yaml
@@ -41,11 +41,11 @@ ai_fairness:
     countries, or patient demographics. Not intended for direct clinical
     decision-making.
 
-  # rai:dataBias
+  # rai:dataBiases
   # Detail specific skews, representational gaps, and known biases — including
   # the geographic or demographic background of annotators, institutional norms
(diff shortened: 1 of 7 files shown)


JoanGi:
The keys are correct! :)!

rajna-fani (approved):
Perfect, now it looks consistent @tompollard Thank you!

PR #76, opened 2026-04-15
Add TSV support and harden CSV type-conflict handling

### LINKED ISSUE #48: Fragility: Dependency on exact PyArrow error message format in CSVHandler
The `CSVHandler` in `src/croissant_maker/handlers/csv_handler.py` uses a regular expression `_ARROW_COL_RE` to parse column indices from PyArrow parsing errors:
```python
_ARROW_COL_RE = re.compile(r"In CSV column #(\d+): CSV conversion error to (\w+)")
```
This regex is highly dependent on the specific error message format of PyArrow. If PyArrow updates its error messages, this regex will fail to match, potentially leading to silent failures or incorrect error reporting to the user. It would be more robust to use structured error information if available from PyArrow or provide a fallback mechanism.

rafiattrach:
## Summary

- Add `TSVHandler` (subclass of `CSVHandler`) for `.tsv`, `.tsv.gz`, `.tsv.bz2`, `.tsv.xz` with correct tab delimiter and `text/tab-separated-values` MIME type
- Fix resource leak in `CSVHandler`: `open_csv` now uses a context manager to guarantee file descriptor release on all runtimes (closes #53)
- Fix fragility on PyArrow error message format: unrecognized conflict messages now trigger a column-probe loop instead of immediately falling back to all-string types (closes #48)
- Promote int/uint type conflicts to `float64` before falling back to `string`, preserving numeric precision (e.g. UniProt `Length`, `Mass` columns)
- Sync `uv.lock` metadata to reflect pyproject.toml constraint bumps from #71–#75

## Test plan

- [ ] `pytest tests/test_csv_handler.py tests/test_tsv_handler.py` — 18 tests pass
- [ ] `pytest tests/test_end_to_end.py::test_uniprot_tsv_generation` — end-to-end TSV test with real UniProt data passes
- [ ] `TSVHandler` does not claim `.csv` files; `CSVHandler` does not claim `.tsv` files
- [ ] fd leak test: 50 sequential schema-only reads complete without resource error
- [ ] Probe fallback test: patching `_parse_conflict` to always return `(None, None)` still produces correct output


src/croissant_baker/handlers/csv_handler.py
rajna-fani, line 1: The unknown-message fallback appears to stringify unresolved columns cumulatively across retries. That can downgrade otherwise healthy columns before the actual conflicting column is reached. Could we adjust this so each probe is non-cumulative and only preserves an override after a successful retry?

tests/test_csv_handler.py
rajna-fani, line 1: This test does not seem to exercise the fallback path yet, since the fixture data here does not create a type conflict. Could we update it to use a dataset that triggers a delayed conflict so the new probe behavior is verified end to end?

rajna-fani (commented):
TSV support looks good overall, but I think the CSV fallback logic still needs another pass. I left some comments below.

rafiattrach:
I removed the probe behavior. I think it adds complexity for a path that main did not rely on, so unknown or unparseable conflict messages now fall back directly to all-string types for correctness.
If PyArrow’s behavior changes in the future, we can revisit that path then or handle it as part of a larger refactor.

Known conflicts still use targeted promotion, and the CSV test now actually exercises the fallback path.
TSV support and the FD fix are unchanged. Full test suite pass.

rajna-fani:
Perfect. I rechecked the latest commit, and the earlier fallback issue now looks fixed. TSV support and the CSV hardening both look good to me, so I think this is good to merge.

rajna-fani (approved):

PR #85, opened 2026-04-16
fix: update README for PyPI page rendering

rafiattrach:
The 0.1.0 wheel uploaded to PyPI was built before the README updates, so the PyPI page shows no description. This bumps a patch version so the next release includes the current README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@
 </p>
 
 <p align="center">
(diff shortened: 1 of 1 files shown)


tompollard (approved):

tompollard (commented):
(for the record, I actually disapprove)

Issue #44, opened 2026-03-29
Improve README documentation and developer experience


smsnobin77:
This issue tracks several documentation and usability improvements to make the repository more accessible for new users and contributors:

1. No minimal working example in the README for first-time users
2. No expected output shown, so users don't know what to expect
3. No progress feedback during long-running dataset processing
4. No contributor guidelines despite the tool's extensible architecture

tompollard:
Thanks @smsnobin77, these points should now be addressed! We have made significant improvements to the codebase and created a new documentation site.

> No minimal working example in the README for first-time users

We provide an example at: https://lcp.mit.edu/croissant-baker/getting-started/

> No expected output shown, so users don't know what to expect

We provide expected output in the example at: https://lcp.mit.edu/croissant-baker/getting-started/

> No progress feedback during long-running dataset processing

Added in: https://github.com/MIT-LCP/croissant-baker/pull/89

> No contributor guidelines despite the tool's extensible architecture

Added guidelines here: https://github.com/MIT-LCP/croissant-baker/blob/main/CONTRIBUTING.md.

tompollard:
@smsnobin77 I think your points are now addressed, but please reopen this issue if not!

PR #89, opened 2026-04-16
feat: add progress bar. ref #44.

tompollard:
As highlighted in https://github.com/MIT-LCP/croissant-baker/issues/44, it would be helpful to provide the user with feedback on progress during creation of Croissant metadata, particularly for large datasets.

This pull request:

- Moves creator parsing and CSV warning logic before the progress context
- Adds per-file progress bar during metadata generation                                                                                                           
- Keeps the spinner for the save/validate phase.

Testing:

- [ ] Run ```croissant-baker --input <dataset_dir>``` on a dataset with multiple files and confirm per-file progress bar is displayed
- [ ] Verify the progress bar shows file count, percentage, and current file name
- [ ] Verify save/validate spinner still works as before

### DIFF
--- a/src/croissant_baker/__main__.py
+++ b/src/croissant_baker/__main__.py
@@ -9,7 +9,13 @@ from pathlib import Path
 from typing import List, Optional
 
 import typer
-from rich.progress import Progress, SpinnerColumn, TextColumn
+from rich.progress import (
+    BarColumn,
+    Progress,
+    SpinnerColumn,
+    TaskProgressColumn,
+    TextColumn,
+)
 
 from croissant_baker.metadata_generator import MetadataGenerator, serialize_datetime
 from croissant_baker.files import discover_files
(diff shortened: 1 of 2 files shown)


rafiattrach (approved):
Nice work! Thanks for taking care of it @tompollard ! I tested this locally on both the bundled MIMIC-IV Demo (32 files) and full MIMIC-IV (~10 GB, 31 compressed CSVs). Progress bar works well, no performance regression, and output is identical.

One thing I noticed: the progress bar total includes discovered files that have no handler (e.g. `index.html`, `README.txt`). On the demo dataset it shows 38/38 but the summary says "Files: 32". Since we don't know which files have handlers until we try each one, the simplest fix would be changing the label from "Processing files..." to "Scanning files..." — that way the count covering all discovered files feels natural.

tompollard:
Thanks Rafi!

> Since we don't know which files have handlers until we try each one, the simplest fix would be changing the label from "Processing files..." to "Scanning files..."

Good catch, I updated the text as suggested.

PR #91, opened 2026-04-19
feat: add DICOM and NIfTI handlers 

varuni7:
Add two new file handlers to support medical and neuroimaging formats:

- DICOMHandler (.dcm, .dicom): reads image geometry, modality, pixel encoding, and acquisition parameters via pydicom (header-only, stop_before_pixels=True). Also detects extension-less DICOM files via magic bytes (DICM at offset 128).

- NIfTIHandler (.nii, .nii.gz): reads spatial dimensions, voxel spacing, data type, NIfTI version, and TR for 4D fMRI volumes via nibabel (header-only, data array never loaded).

Both handlers implement build_croissant() to produce a FileSet and summary RecordSet. The tr_seconds field is only added when at least one 4D volume is present in the dataset.

Also adds:
- pydicom>=2.4 and nibabel>=5.0 to dependencies
- 48 tests (24 per handler) using tmp_path + synthetic files
- DICOM and NIfTI sections in supported-formats.md
- Regenerated docs/_generated/formats-table.md
- Updated E2E snapshot outputs from upstream mlcroissant changes


src/croissant_baker/handlers/dicom_handler.py
```diff
@@ -0,0 +1,284 @@
...
+            return default
+        if hasattr(val, "__iter__") and not isinstance(val, (str, bytes)):
+            return [float(v) for v in val]
+        return val
+    except Exception:
+        return default
+
+
+def _read_dicom_properties(file_path: Path) -> Dict:
+    import pydicom
```
tompollard: Please could you move this import to the top of the module?

src/croissant_baker/handlers/nifti_handler.py
```diff
@@ -0,0 +1,300 @@
...
+
+
+def _read_nifti_properties(file_path: Path) -> Dict:
+    """
+    Read NIfTI metadata using nibabel (header only, data array never loaded).
+
+    Extracts spatial dimensions, voxel spacing, data type, and TR for 4D volumes.
+    Works for both NIfTI-1 (.nii, .nii.gz) and NIfTI-2 formats.
+    """
+    import nibabel as nib
```
tompollard: Please move import to top of module

tests/data/output/eicu_demo_croissant.jsonld
```diff
@@ -68,7 +69,7 @@
       "name": "Jesse Raffa"
     }
   ],
-  "datePublished": "2021-05-06T00:00:00",
+  "datePublished": "2021-05-06",
```
tompollard: Why has this changed? (`"datePublished": "2021-05-06T00:00:00"` -> `"datePublished": "2021-05-06"`)

tests/data/output/gharchive_demo_croissant.jsonld
```diff
@@ -54,7 +55,7 @@
     "@type": "sc:Person",
     "name": "GitHub Archive"
   },
-  "datePublished": "2023-01-01T00:00:00",
+  "datePublished": "2023-01-01",
```
tompollard: Why has this changed? (`"datePublished": "2023-01-01T00:00:00"` -> `"datePublished": "2023-01-01"`)

tests/data/output/glaucoma_fundus_croissant.jsonld
```diff
@@ -54,7 +55,7 @@
     "@type": "sc:Person",
     "name": "Hillel Yaffe Medical Center"
   },
-  "datePublished": "2024-11-14T00:00:00",
+  "datePublished": "2024-11-14",
```
tompollard: As above, why has this timestamp changed?

tests/data/output/mimiciv_demo_croissant.jsonld
```diff
@@ -85,7 +86,7 @@
       "url": "https://lcp.mit.edu/"
     }
   ],
-  "datePublished": "2023-01-06T00:00:00",
+  "datePublished": "2023-01-06",
```
tompollard: As above, why has this timestamp changed?

tests/test_nifti_handler.py
```diff
@@ -0,0 +1,326 @@
...
+    assert "tr_seconds" in field_names
+
+
+# ---------------------------------------------------------------------------
+# Handler registration
+# ---------------------------------------------------------------------------
+
+
+def test_nifti_handler_registered() -> None:
+    from croissant_baker.handlers.registry import find_handler, register_all_handlers
```
tompollard: Unless there is a good reason to import within a function, please move imports to the top of the module.

tompollard (commented):
Thanks @varuni7, this is great. I have added a few small notes inline, and will do some testing.

tompollard:
I tested on several DICOMs downloaded from PhysioNet and overall it worked well. One minor issue is that a couple of the DICOMs had empty modality (0008,0060) fields in their DICOM header, so the handler's `_safe_get(ds, "Modality")` call returns `None`. 

As a result, the fileset description shows `6 DICOM file(s) (DX (2), US (2))` (i.e. the modalities sum to 4 but there are 6 files). 

```
    {
      "@type": "cr:FileSet",
      "@id": "dicom-files",
      "name": "DICOM files",
      "description": "6 DICOM file(s) (DX (2), US (2))",
      "encodingFormat": "application/dicom",
      "includes": [
        "**/*.dcm",
        "**/*.dicom"
      ]
    }
```

Perhaps instead we should acknowledge the missing modalities here (e.g. `6 DICOM file(s) (DX (2), US (2), unknown (2))`)?


Another minor question is whether we should capture DICOM tags in the metadata to demonstrate where values were extracted from, e.g. instead of:

```
  {                                                                                                                                               
    "@type": "cr:Field",
    "@id": "dicom/modality",                                                                                                                      
    "name": "modality",                                                                                                                           
    "description": "DICOM modality (e.g. CT, MR, PT)",                                                                                            
    "dataType": "sc:Text",
    "source": {
      "fileSet": { "@id": "dicom-files" },
      "extract": { "fileProperty": "content" }
    }
  }

```

...we could do:

```
  {
    "@type": "cr:Field",
    "@id": "dicom/modality",
    "name": "modality",
    "description": "DICOM Modality (0008,0060)",
    "dataType": "sc:Text",
    "source": {
      "fileSet": { "@id": "dicom-files" },
      "extract": { "fileProperty": "content" }
    }
  }
```

These points don't need to be addressed in this PR. We could raise an issue and add in a future improvement.

tompollard:
Also tested on several .nii files and it worked well. As with DICOM, it would be nice if we could capture which values were extracted from which NIfTI header property. This could be addressed in a later PR.

src/croissant_baker/handlers/dicom_handler.py
```diff
@@ -0,0 +1,284 @@
...
+
+    ds = pydicom.dcmread(str(file_path), stop_before_pixels=True)
```
rafiattrach, line 45: Tested this on pydicom's own test corpus (`pydicom/pydicom-data`, 85 files). 5 files fail with `'DICM' prefix is missing from the header. Use force=True to force reading.` These are raw pixel stream fragments referenced from a `DICOMDIR`, not broken DICOMs. Per DICOM Part 10 a file without the DICM preamble at offset 128 is not a standalone DICOM file, so the correct behaviour is to skip, not raise.

Suggested fix: run `.dcm` and `.dicom` through the existing `_has_dicom_magic` guard (currently only used on the extensionless branch) and skip cleanly when the preamble is missing. A single summary line like `Skipped N DICOM file(s) without DICM preamble` keeps the output honest without drowning the user in per file tracebacks. Pairs with the `can_handle` comment below.

src/croissant_baker/handlers/dicom_handler.py
```diff
@@ -0,0 +1,284 @@
...
+
+    EXTENSIONS = (".dcm", ".dicom")
+    FORMAT_NAME = "DICOM"
+    FORMAT_DESCRIPTION = (
+        "Image geometry, modality, pixel encoding, acquisition parameters"
+    )
+
+    def can_handle(self, file_path: Path) -> bool:
+        suffix = file_path.suffix.lower()
+        if suffix in SUPPORTED_EXTENSIONS:
```
rafiattrach: `can_handle` defines the handler's contract to the registry, so when it says `True` and then `extract_metadata` raises on the file's content, the contract is broken. Right now a `.dcm` file without the DICM preamble returns `True` here but then raises inside `extract_metadata`, while an extensionless file with identical content is correctly rejected. The asymmetry has no principled basis.

I had a look at the other handlers to see whether this is a broader concern. Text format handlers (CSV, TSV, JSON, FHIR, WFDB `.hea`) are fine as they are because content validation is essentially parsing. The handlers that would benefit from the same uniform magic byte check are the other binary formats:

- `ImageHandler`: PNG (`\x89PNG`), JPEG (`\xff\xd8\xff`), TIFF (`II*\0` / `MM\0*`) all have well known magic bytes
- `ParquetHandler`: `PAR1` at the start and end of every Parquet file
- `NIfTIHandler`: `n+1\0` at offset 344 (NIfTI-1) or `n+2\0` at offset 4 (NIfTI-2)

Not suggesting we fix those in this PR, that would blow the scope. But it would be good to fix the DICOM asymmetry here, since it is already half implemented, and open a follow up issue to extend the same pattern to the other binary handlers. Suggested fix for this PR: apply `_has_dicom_magic` to `.dcm` and `.dicom` too, not only the extensionless branch, and skip cleanly when the preamble is missing (pairs with the `_read_dicom_properties` comment above).

src/croissant_baker/handlers/dicom_handler.py
```diff
@@ -0,0 +1,284 @@
...
+
+    series_desc = _safe_get(ds, "SeriesDescription")
+    if series_desc is not None:
+        props["series_description"] = str(series_desc).strip()
+
+    manufacturer = _safe_get(ds, "Manufacturer")
+    if manufacturer is not None:
+        props["manufacturer"] = str(manufacturer).strip()
+
+    sop_class = _safe_get(ds, "SOPClassUID")
```
rafiattrach, line 98: This captures `SOPClassUID` but not `(0020,000D) StudyInstanceUID` or `(0020,000E) SeriesInstanceUID`. 

From what I researched, DICOM organises data as a 4 level tree: Patient, Study, Series, Instance. Each level has its own UID. `SeriesInstanceUID` is the one that groups images into scans (one CT series is one scan, one MR series is one pulse sequence). `SOPClassUID` is a different thing: it is a class reference that tells you what *kind* of object the file is ("CT Image Storage", "MR Image Storage", "Segmentation Storage"), not which scan it belongs to.

Without `SeriesInstanceUID` in the emitted metadata, a user with 500 CT slices on disk cannot regroup them into the 20 scans they belong to, which is the most basic operation anyone performs on a DICOM dataset. This is a functional gap rather than a nice to have.

Let's add both if you agree, exposed as `cr:Field`s in `build_croissant`. Something like this?

```python
study_uid = _safe_get(ds, "StudyInstanceUID")
if study_uid is not None:
    props["study_instance_uid"] = str(study_uid)

series_uid = _safe_get(ds, "SeriesInstanceUID")
if series_uid is not None:
    props["series_instance_uid"] = str(series_uid)
```

src/croissant_baker/handlers/nifti_handler.py
```diff
@@ -0,0 +1,300 @@
...
+        t_range = summary.get("dim_t_range")
+        ndim_max = summary.get("ndim_max", 3)
+        if ndim_max >= 4 and t_range:
+            dims_note += f", {t_range[0]}-{t_range[1]} volumes"
+
+        dtype_counts = summary.get("dtype_counts", {})
+        dtypes_str = ", ".join(dtype_counts.keys()) if dtype_counts else "unknown dtype"
+
+        mime_types = set()
+        includes = []
```
rafiattrach: `includes = []` is overwritten four lines later and does not depend on per file metadata

could be removed/tidied up?

src/croissant_baker/handlers/nifti_handler.py
```diff
@@ -0,0 +1,300 @@
...
+            r = summary.get(key)
+            if r:
+                dims_parts.append(f"{r[0]}" if r[0] == r[1] else f"{r[0]}-{r[1]}")
+        dims_note = "x".join(dims_parts) if dims_parts else "unknown dims"
+
+        # 4D note.
+        t_range = summary.get("dim_t_range")
+        ndim_max = summary.get("ndim_max", 3)
+        if ndim_max >= 4 and t_range:
+            dims_note += f", {t_range[0]}-{t_range[1]} volumes"
```
rafiattrach: The spatial dim loop above already collapses `N-N` to `N`. Same rule here keeps the description consistent so a homogeneous 4D set reads `(64x64x30, 100 volumes)` instead of `(64x64x30, 100-100 volumes)`, potentially something like this:

```python
if ndim_max >= 4 and t_range:
    if t_range[0] == t_range[1]:
        dims_note += f", {t_range[0]} volumes"
    else:
        dims_note += f", {t_range[0]}-{t_range[1]} volumes"
```

rafiattrach (changes requested):
@varuni7 Thanks for the great addition! Could you please take a look at some of the comments down below?

src/croissant_baker/handlers/dicom_handler.py
varuni7: agreed

tests/data/output/mimiciv_demo_croissant.jsonld
varuni7: The date format change came from running the E2E tests locally after the pull, a newer version of mlcroissant in the local venv serialises date-only values without T00:00:00, which caused the snapshots to diff.
reverting it

rafiattrach:
@varuni7 Thanks! imports, snapshots, and the `.dcm` magic check look good.

A few things still open from the earlier round. Happy to take it the rest of the way if it's easier on your end, otherwise are you able to get to them in the next day or so?

- modality counter undercounts when files have no Modality tag (e.g. output reads `4 file(s) (MR (1), CT (1), US (1))`)
- on a real-world set (pydicom-data) a handful of valid DICOMs surface as exceptions during the run instead of skipping cleanly
- nothing captures `PatientID` / `StudyInstanceUID` / `SeriesInstanceUID`, which are the natural grouping keys for a DICOM collection
- minor: dead lines around `nifti_handler.py:151-155`, and the "N-N volumes" reading when min == max

Let me know which way works for you!

varuni7:
> @varuni7 Thanks! imports, snapshots, and the `.dcm` magic check look good.
> 
> A few things still open from the earlier round. Happy to take it the rest of the way if it's easier on your end, otherwise are you able to get to them in the next day or so?
> 
> * modality counter undercounts when files have no Modality tag (e.g. output reads `4 file(s) (MR (1), CT (1), US (1))`)
> * on a real-world set (pydicom-data) a handful of valid DICOMs surface as exceptions during the run instead of skipping cleanly
> * nothing captures `PatientID` / `StudyInstanceUID` / `SeriesInstanceUID`, which are the natural grouping keys for a DICOM collection
> * minor: dead lines around `nifti_handler.py:151-155`, and the "N-N volumes" reading when min == max
> 
> Let me know which way works for you!

Hi @rafiattrach!

Thanks for the feedback! I won't be able to get to these specific points until the weekend, so I’d be more than happy to collaborate on this.

Please feel free to take it the rest of the way!

rafiattrach:
@varuni7 @tompollard thanks for PR & the review!

Pushed a follow-up addressing the open comments and consolidating the SPECT demo into one mixed-format fixture. Local & CI green and re-verified on the [pydicom test corpus](https://github.com/pydicom/pydicom-data), where the original DICM-preamble issue surfaced.

Fixture data: open-access [PhysioNet Myocardial Perfusion SPECT database](https://physionet.org/content/myocardial-perfusion-spect/1.0.0/)

PR #94, opened 2026-04-25
fix: count only cr:FileObject entries in CLI summary

### LINKED ISSUE #92: CLI "Files" count in summary includes FileSets, not just FileObjects
When `croissant-baker` finishes a run it prints a summary banner with `Files: N`. That number is computed from the length of the generated Croissant `distribution` array, but `distribution` holds both `cr:FileObject` entries (one per processed file) and `cr:FileSet` entries (one per handler that groups files, like "DICOM files" or "NIfTI files"). Every handler that emits a FileSet inflates the reported file count by one.

Concrete example on an 85 file DICOM dataset where 5 files fail: the DICOM FileSet description (emitted by the handler) correctly says `80 DICOM file(s)`. The CLI banner says `Files: 81`. The handler uses `len(file_metas)`, the banner uses `len(distribution)`, and they drift by the number of FileSet entries.

Potential fix is a one liner in the CLI summary path: filter to `cr:FileObject` when counting:

```python
file_count = sum(
    1 for d in metadata_dict.get("distribution", [])
    if d.get("@type") == "cr:FileObject"
)
```

Affected handlers (any that emit FileSets): Image, Parquet, FHIR, JSON, WFDB, DICOM, NIfTI. The discrepancy has likely existed for a while but became visible during review of PR #91. Worth a small PR that also adds a test asserting the banner count matches the number of processed files.

rafiattrach:
The Files line used len(distribution), which mixes FileObject and FileSet entries and inflated the count for partitioned Parquet and image datasets. Generate and validate paths now filter on @type, and a File sets line is emitted when FileSets are present.

Closes #92

### DIFF
--- a/src/croissant_baker/__main__.py
+++ b/src/croissant_baker/__main__.py
@@ -118,6 +118,13 @@ _SPEC_REQUIRED_FLAGS = {
 _RAI_CONFORMS_TO = "http://mlcommons.org/croissant/RAI/1.0"
 
 
+def _echo_file_counts(file_count: int, file_set_count: int) -> None:
+    """Print Files and File sets banner lines (File sets only when present)."""
+    typer.echo(f"Files: {file_count}")
+    if file_set_count:
+        typer.echo(f"File sets: {file_set_count}")
+
+
 def _warn_missing_spec_fields(**provided: object) -> None:
     """Warn about spec-required fields that were not explicitly provided."""
     missing = [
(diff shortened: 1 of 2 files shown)


rajna-fani (approved):
Fixes the inflated file count correctly in both the generate and validate paths, with good test coverage. All CI passes. Looks good to me!

PR #95, opened 2026-04-26
fix: validate magic bytes in image and parquet can_handle (#93)

### LINKED ISSUE #93: Apply magic byte validation in can_handle for binary format handlers
Each handler exposes `can_handle(file_path)` which the registry uses to dispatch files. The implicit contract is: if `can_handle` returns True, `extract_metadata` will not raise on the file's content. Right now most binary format handlers violate this contract because they dispatch on extension only. A `.png` file that is actually HTML, or a truncated `.parquet` file, will be claimed by the handler and then crash inside `extract_metadata` with a library traceback. On real datasets the result is noisy per file warnings and reduced trust in the registry's dispatch decision.

Text format handlers (CSV, TSV, JSON/JSONL, FHIR/NDJSON, WFDB `.hea`) can stay as they are because content validation would amount to parsing. The handlers that should adopt a cheap magic byte check:

| Handler | Magic bytes |
|---|---|
| Image | PNG `\x89PNG\r\n\x1a\n`, JPEG `\xff\xd8\xff`, TIFF `II*\0` or `MM\0*`, GIF `GIF87a` or `GIF89a` |
| Parquet | `PAR1` at file start and end |
| NIfTI | `n+1\0` at offset 344 (NIfTI-1), `n+2\0` at offset 4 (NIfTI-2) |

The DICOM handler already has `_has_dicom_magic` and is being fixed in PR #91 to apply it uniformly across extensionless and extensioned files. The same pattern can be reused: a small helper per handler that reads a handful of bytes and returns a bool, invoked from `can_handle`. Files that fail the check are silently skipped by the registry rather than surfacing a traceback per file.

Acceptance: each listed handler's `can_handle` returns False when the file lacks the expected magic bytes. Each handler adds one or two unit tests covering a file with the right extension but wrong content.

rafiattrach:
The handler registry contract was implicit: when can_handle returns True, extract_metadata should not raise on the file's content. Image and parquet handlers dispatched on extension alone, so renamed or truncated files crashed inside Pillow or pyarrow with library tracebacks.

This change adds a magic byte check to can_handle for both handlers. Image covers PNG, JPEG, GIF, TIFF (including BigTIFF), BMP, WebP, ICO, each signature cited from the format spec. Parquet requires PAR1 at start AND end, so truncated files are rejected. Files that fail the check are skipped by the registry and a WARNING is logged that names the file and the specific reason (wrong magic, missing footer, too small) so users see what was dropped instead of being blindsided by a smaller file count.

Verified on MIMIC IV MEDS (369 parquet files), MIMIC IV CSV (31 files), and a mixed impostor fixture where four bad files were dropped with clear stderr warnings naming each one. 189 tests pass.

Closes #93.

### DIFF
--- a/src/croissant_baker/handlers/image_handler.py
+++ b/src/croissant_baker/handlers/image_handler.py
@@ -41,6 +41,68 @@ _MIME_TYPES: Dict[str, str] = {
     ".tif": "image/tiff",
 }
 
+# Magic-byte signatures for every supported image extension. These are
+# stable, format-spec-defined headers — they don't change across versions:
+#
+#   PNG     : ISO/IEC 15948 §5.2 — 8-byte signature
+#   JPEG    : ITU-T T.81 / JFIF  — SOI marker 0xFFD8 followed by another marker (0xFF**)
+#   GIF     : GIF89a spec        — ASCII "GIF87a" or "GIF89a"
+#   TIFF    : TIFF 6.0 §2        — "II\x2a\x00" (LE) or "MM\x00\x2a" (BE);
+#                                  BigTIFF (Adobe ext, 2007) uses version byte
+#                                  0x2b instead of 0x2a — read by Pillow and tifffile
+#   BMP     : BITMAPFILEHEADER   — bfType "BM"
+#   WebP    : RFC 6386           — "RIFF" + 4-byte size + "WEBP"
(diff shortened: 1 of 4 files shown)


rajna-fani (approved):
LGTM. The magic-byte checks are spec-correct for both handlers, the logging contract is consistent, and the test coverage looks solid. One gap from #93: NIfTI still routes by extension alone. Worth a follow-up PR so the registry contract holds for all binary handlers.

PR #101, opened 2026-05-03
Fix/id collisions followup #63

rajna-fani:
When two files share the same basename in different subdirectories, or two column names sanitize to the same    
  string, the generator was emitting duplicate @id values. JSON-LD consumers silently merge nodes that share an
  @id, causing missing data or cryptic validation errors.                                                         
                                                             
  This adds disambiguation helpers in utils.py that prefix colliding identifiers with parent directory components,
   updates all handlers to use them, and adds a generator-level uniqueness check that catches any future          
  regressions before serialization.

Follow-up to PR #63


rafiattrach:
good catch! just please fix CI, its a small pre-commit probably

PR #111, opened 2026-06-30
fix: make generated metadata reproducible (drop wall-clock defaults)

elementalcollision:

## Problem: output is not reproducible

`MetadataGenerator` derived parts of its output from `datetime.now()` whenever the caller didn't supply a date or citation. Baking the **same dataset twice, one second apart**, produces two different files:

```
$ croissant-baker --input demo/ --creator "T,t@e.com" --name X -o run1.jsonld
$ croissant-baker --input demo/ --creator "T,t@e.com" --name X -o run2.jsonld
$ diff run1.jsonld run2.jsonld
<   "datePublished": "2026-06-30T12:36:52.236716"
>   "datePublished": "2026-06-30T12:36:53.177421"
```

Two sources of wall-clock leakage:

1. **`datePublished`** defaulted to `datetime.now()` — the metadata-generation timestamp, down to the microsecond. That breaks reproducibility (spurious VCS diffs, broken content-addressed caching) and is semantically wrong: the dataset was not *published* at generation time.
2. **The default citation** embedded `datetime.now().year` *and* a fabricated author — `"Dataset Creator. (2026). <name> Dataset. Generated with automated type inference."` — even when real `--creator` values were supplied. (You can see this today in 7 committed example outputs under `tests/data/output/`, e.g. the MEDS one reads `"Dataset Creator. (2026). …"`.)

## Fix — never derive output from wall-clock state

- **`_resolve_date()`** omits `datePublished` when none is supplied (it's optional in schema.org/Croissant) instead of stamping `now()`.
- **`_build_citation()`** synthesises the default citation from the supplied **creators** and the **publication/creation-date year**; if neither is available there's nothing real to cite, so `cite_as` is omitted rather than invented.

## Tests & artifacts

- New `tests/test_determinism.py`: bake-twice byte-equality, `datePublished` omission, citation derivation, and that omitting `datePublished` still passes `mlcroissant` validation.
- Full suite: **274 passed** (269 + 5 new).
- Regenerated the 7 committed example outputs — each is a **one-line** `citeAs` change, e.g.:
  ```
  - "Dataset Creator. (2026). MIMIC-IV demo … Dataset. Generated with automated type inference."
  + "Robin van de Water, Matthew McDermott. (2025). MIMIC-IV demo … [Data set]."
  ```

## Behavior change (intentional)

Default `datePublished` and the default `citeAs` format change for callers who don't pass `--date-published` / `--citation`. This is the point — the previous defaults fabricated metadata and drifted with the clock. Callers who pass those flags explicitly are unaffected. Happy to adjust the default citation wording (`… [Data set].`) to your preference.

### DIFF
--- a/src/croissant_baker/metadata_generator.py
+++ b/src/croissant_baker/metadata_generator.py
@@ -467,16 +467,58 @@ class MetadataGenerator:
                 f"Expected '2023-12-15' or '2023-12-15T10:30:00'. Error: {e}"
             )
 
-    def _build_citation(self) -> str:
+    def _build_citation(self) -> Optional[str]:
+        """Return a deterministic default citation, or None.
+
+        A caller-supplied ``citation`` is used verbatim. Otherwise we synthesise
+        one only from metadata we actually have — never from wall-clock state, so
+        the same dataset always bakes to the same citation. The year comes from a
+        supplied publication/creation date (the old ``datetime.now().year`` made
+        output drift across year boundaries); the author comes from the supplied
+        creators (the old default fabricated "Dataset Creator" even when real
+        creators were given). If neither a creator nor a date is available there
+        is nothing real to cite, so we omit ``cite_as`` rather than invent it.
+        """
         if self.citation:
             return self.citation
-        year = datetime.now().year
-        name = self.name or self.dataset_path.name
-        return f"Dataset Creator. ({year}). {name} Dataset. Generated with automated type inference."
 
-    def _resolve_date(self) -> datetime:
+        names = [
+            c["name"]
+            for c in (self.creators or [])
+            if isinstance(c, dict) and c.get("name")
+        ]
+        author = ", ".join(names) if names else None
+
+        year = None
+        for raw in (self.date_published, self.date_created):
+            if not raw:
+                continue
+            try:
+                year = datetime.fromisoformat(raw).year
+                break
+            except ValueError:
+                continue
+
+        if author is None and year is None:
+            return None
+
+        name = self.name or self.dataset_path.name
+        author_part = author if author else "Unknown"
+        year_part = f" ({year})." if year is not None else ""
+        return f"{author_part}.{year_part} {name} [Data set]."
+
+    def _resolve_date(self) -> Optional[datetime]:
+        """Return the parsed publication date, or None when none was supplied.
+
+        Previously this defaulted to ``datetime.now()``, stamping the
+        metadata-generation time (down to the microsecond) as the dataset's
+        ``datePublished``. That made output non-reproducible — baking the same
+        dataset twice produced two different files — and is semantically wrong,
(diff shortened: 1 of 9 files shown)


src/croissant_baker/metadata_generator.py
```diff
@@ -467,16 +467,58 @@ def _parse_iso(value: Optional[str]):
                 f"Expected '2023-12-15' or '2023-12-15T10:30:00'. Error: {e}"
             )
 
-    def _build_citation(self) -> str:
+    def _build_citation(self) -> Optional[str]:
+        """Return a deterministic default citation, or None.
```
rafiattrach: This docstring describes code that no longer exists ("the old `datetime.now().year`",
"the old default fabricated 'Dataset Creator'"). That history is great in the PR, but
in the source it's noise to a future reader. Present tense, keep the *why* maybe something like this:

    """Build the default citation, or None.

    A caller-supplied citation is used verbatim. Otherwise it's derived only from
    real metadata (creators and the publication/creation year), so the same input
    always bakes to the same citation. 
    """

src/croissant_baker/metadata_generator.py
```diff
@@ -467,16 +467,58 @@ def _parse_iso(value: Optional[str]):
...
+        if author is None and year is None:
+            return None
+
+        name = self.name or self.dataset_path.name
+        author_part = author if author else "Unknown"
+        year_part = f" ({year})." if year is not None else ""
+        return f"{author_part}.{year_part} {name} [Data set]."
+
+    def _resolve_date(self) -> Optional[datetime]:
+        """Return the parsed publication date, or None when none was supplied.
```
rafiattrach, line 508: Same: "Previously this defaulted to `datetime.now()`..." narrates the old bug.

rafiattrach (changes requested):
Thanks! Some minor comments:`test_determinism.py` covers "creator + date" and "neither," but not the
creator-without-date case (`"Name. Title [Data set]."`) — which is the common CLI
path since `--creator` is required but `--date-published` is optional. Worth pinning.

elementalcollision:
Thanks for the careful review, @rafiattrach — all three points addressed in b8fcc9a:

1. **Creator-without-date test** — added `test_default_citation_creator_without_date`, pinning the common CLI path (`--creator` required, `--date-published` optional) to `"Name. Title [Data set]."` with no year. That was indeed the one default-citation branch the suite didn't cover.
2. **`_build_citation` docstring** — rewritten in present tense; keeps the determinism *why*, drops the `datetime.now().year` / "Dataset Creator" narration (that history lives in the PR description now).
3. **`_resolve_date` docstring** — same treatment: states that `datePublished` is omitted when unset (keeps output reproducible; optional in schema.org/Croissant), no "previously…".

Full suite green (275). Ready for another look whenever you have a moment — thanks again.

PR #113, opened 2026-06-30
fix: bound discovery memory and route generator warnings through logging

elementalcollision:

Two small robustness/consistency fixes, bundled.

## 1. Bounded-memory file discovery

`files.discover_files` appended **every** hidden-directory path to `skipped_examples`, a list used only to print a few examples in a debug log. On a dataset with a large hidden tree — an accidental `.git`, a big `.ipynb_checkpoints` — that list grows without bound for output we only ever sample. Cap retained examples at 5; the skipped **count** stays exact.

## 2. Consistent observability (logging, not print)

Every file handler already logs through a module logger (`croissant_baker.handlers.*`), but `metadata_generator` emitted its warnings via `print()` to **stdout**. This routes the three warning-level diagnostics — failed extraction, failed `build_croissant`, multi-match field mapping — through a `croissant_baker.metadata_generator` logger: off stdout, controllable, and consistent with the handlers.

The DICOM skip summary deliberately **stays** a stdout `print` — it's a user-facing summary line with its own dedicated test (`test_dicom_skip_summary_printed_…`), not a warning.

## Tests

- New cap assertion in `test_files.py` (20 hidden files → exact count, 5 examples).
- New `test_generator_diagnostics.py`: the field-mapping warning is emitted on the logger.
- The existing field-mapping CLI test now asserts via `caplog` — the same pattern the parquet/image/fhir handler-warning tests already use.
- Full suite: **271 passed**. No change to committed `tests/data/output/*.jsonld`.

## Note on overlap

This touches the same `metadata_generator.py` "Failed to process" warning line that #112 (parallel extraction) relocates. The two overlap by one line; whichever merges second needs a trivial rebase.  The changes are otherwise independent.

### DIFF
--- a/src/croissant_baker/files.py
+++ b/src/croissant_baker/files.py
@@ -6,6 +6,12 @@ from typing import List, Optional
 
 logger = logging.getLogger(__name__)
 
+# Cap on example paths retained for the hidden-directory skip debug log. The
+# skipped count is always exact; only this example list is bounded, so a dataset
+# with a huge hidden tree (an accidental .git or .ipynb_checkpoints, say) cannot
+# accumulate an unbounded list of paths we only ever sample from.
+_MAX_SKIPPED_EXAMPLES = 5
+
 
 def discover_files(
     dir_path: str,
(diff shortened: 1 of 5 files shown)


rafiattrach (approved):
Thanks! That seems reasonable!

PR #114, opened 2026-06-30
feat: detect foreign keys and emit cr:references between RecordSets

elementalcollision:
Draft PR opened by **Chimera** (an autonomous code agent operated by @elementalcollision), as part of a review of croissant-baker. Opened as a **draft** for maintainer review — especially on the design choice flagged at the bottom.

Addresses **#51** (and the inline `# TODO … references: detect foreign-key columns` in `generate_metadata`).

## What

Today every file becomes an isolated RecordSet; relational datasets lose their join structure. This adds an **opt-in** detector — `--detect-references` (`detect_references=True`) — that links shared key columns across RecordSets with Croissant `cr:references`.

It is deliberately conservative and **never guesses a direction**:

- A column is a foreign-key candidate when it's named `<stem>_id` and appears in **≥2 RecordSets**. A bare `id` is excluded (too generic).
- The **parent** is the RecordSet named after the stem (singular/plural: `study_id` → `studies`/`study`) that also carries the key column. Every other holder's field gets `references → parent.<stem>_id`.
- Shared keys with **no name-identifiable parent** are *logged as unresolved, not linked* — surfacing the candidate without inventing a relationship that might point the wrong way.

Example (`studies.csv` + `samples.csv`, both with `study_id`):

```json
"field": [
  { "@id": "samples/study_id", "name": "study_id",
    "references": { "field": { "@id": "studies/study_id" } } }
]
```

## Design

- Detection is a **pure function** in a new `references.py` (trivially unit-testable); `metadata_generator` maps its output onto the real `Field` objects via the native `mlc.Source(field=…)` API — no JSON patching.
- **Off by default** — existing output is unchanged; the committed `tests/data/output/*.jsonld` are untouched.

## Tests

`tests/test_references.py`: detector unit tests (same-name match, `y→ies` plural, unresolved reporting, rejection of non-key / unshared / bare-`id` columns) and end-to-end (emission, **mlcroissant validation**, opt-in default, determinism). Full suite: **279 passed**. ruff clean.

## Open question for maintainers (why it's a draft)

This v1 picks the parent **by name**. That cleanly handles the same-named-key convention #51 describes (e.g. OMOP/MEDS-style repeated `subject_id`), but it will **not** link cases where the parent's name doesn't match the stem, notably MIMIC's `subject_id → patients` (there's no `subjects` table). The robust fix is to pick the parent by **column-value uniqueness** (the actual primary-key signal), which would catch those at the cost of reading the candidate key columns. I scoped v1 to the safe, no-extra-IO heuristic on purpose; happy to extend to uniqueness-based detection if you'd prefer that direction.

### DIFF
--- a/docs/reference/api.md
+++ b/docs/reference/api.md
@@ -12,6 +12,7 @@ programmatically — without the CLI.
         - generate_metadata
         - save_metadata
         - scan_report
+        - reference_report
 
 ## Scan coverage
 
@@ -34,6 +35,22 @@ still explain itself.
 ::: croissant_baker.entries.Outcome
 
 ::: croissant_baker.entries.Reason
+
+## Foreign-key detection
+
+Opt-in, via `--detect-references` or `detect_references=True`. The report is
+`None` when the pass did not run, so "found nothing" and "never asked" stay
+distinguishable. Shared key columns it declines to link are carried here rather
+than dropped.
+
+::: croissant_baker.references.ReferenceReport
+    options:
+      members:
+        - links
+        - unresolved
+        - summary_lines
+
+::: croissant_baker.references.Unlinkable
 
 ## File Discovery
 
--- a/docs/reference/cli.md
+++ b/docs/reference/cli.md
@@ -40,6 +40,7 @@ $ croissant-baker [OPTIONS] COMMAND [ARGS]...
 * `--field-mapping TEXT`: Link one column to an external vocabulary URI. Format: 'COLUMN=URI'. Example: --field-mapping 'age=http://www.wikidata.org/entity/Q11464'. Matches by bare column name across all RecordSets; a warning prints if a name resolves to multiple fields. Repeatable; combine with --field-mappings (flags override YAML).
 * `--count-csv-rows`: Count exact row numbers for CSV files (slow for large datasets)
 * `-j, --jobs INTEGER`: Worker threads for file extraction. 0 = auto (from CPU count), 1 = serial. Output is identical regardless of this value.  [default: 0]
+* `--detect-references`: Detect foreign keys between tables that share a key column (e.g. subject_id) and emit cr:references links. Conservative: links only when a parent table is identifiable by name; shared keys it will not link are reported, and named under --verbose.
 * `--rai-data-collection TEXT`: How and where the data was gathered.
 * `--rai-data-collection-type TEXT`: Collection type, e.g. 'observational'. Can be used multiple times.
 * `--rai-data-collection-missing-data TEXT`: How missing data was handled during collection.
--- a/src/croissant_baker/__main__.py
+++ b/src/croissant_baker/__main__.py
@@ -501,6 +501,11 @@ def main(
         "-j",
         help="Worker threads for file extraction. 0 = auto (from CPU count), 1 = serial. Output is identical regardless of this value.",
     ),
+    detect_references: bool = typer.Option(
+        False,
+        "--detect-references",
+        help="Detect foreign keys between tables that share a key column (e.g. subject_id) and emit cr:references links. Conservative: links only when a parent table is identifiable by name.",
+    ),
     # Native mlcroissant RAI fields exposed directly as CLI flags.
     rai_data_collection: Optional[str] = typer.Option(
         None, "--rai-data-collection", help="How and where the data was gathered."
@@ -794,6 +799,7 @@ def main(
             field_mappings=merged_field_mappings,
             count_csv_rows=count_csv_rows,
             max_workers=jobs or None,
+            detect_references=detect_references,
             includes=include,
             excludes=exclude,
             rai_fields=native_rai_fields,
--- a/src/croissant_baker/__main__.py
+++ b/src/croissant_baker/__main__.py
@@ -178,6 +178,25 @@ def _echo_scan_coverage(
         )
 
 
+def _echo_reference_summary(
+    generator: Optional["MetadataGenerator"], verbose: bool
+) -> None:
+    """Print what foreign-key detection linked, and what it declined to link.
+
+    Silent unless ``--detect-references`` ran: the report is ``None`` otherwise,
+    so a default bake says nothing about a pass it never made. Like the coverage
+    section, the default is one line whatever the dataset, and only ``--verbose``
+    names a column per line.
+    """
+    if generator is None:
+        return
+    report = generator.reference_report
+    if report is None:
+        return
+    for line in report.summary_lines(verbose=verbose):
+        typer.echo(line)
+
+
 def _warn_missing_spec_fields(**provided: object) -> None:
     """Warn about spec-required fields that were not explicitly provided."""
     missing = [
@@ -554,7 +573,7 @@ def main(
     detect_references: bool = typer.Option(
         False,
         "--detect-references",
-        help="Detect foreign keys between tables that share a key column (e.g. subject_id) and emit cr:references links. Conservative: links only when a parent table is identifiable by name.",
+        help="Detect foreign keys between tables that share a key column (e.g. subject_id) and emit cr:references links. Conservative: links only when a parent table is identifiable by name; shared keys it will not link are reported, and named under --verbose.",
     ),
     # Native mlcroissant RAI fields exposed directly as CLI flags.
     rai_data_collection: Optional[str] = typer.Option(
@@ -961,6 +980,7 @@ def main(
         _echo_file_counts(file_count, file_set_count)
         typer.echo(f"Record sets: {record_count}")
         _echo_scan_coverage(generator, report, verbose)
+        _echo_reference_summary(generator, verbose)
         typer.echo(f"Saved to: {output}")
 
         if not validate:
--- a/src/croissant_baker/__main__.py
+++ b/src/croissant_baker/__main__.py
@@ -551,6 +551,11 @@ def main(
         "-j",
         help="Worker threads for file extraction. 0 = auto (from CPU count), 1 = serial. Output is identical regardless of this value.",
     ),
+    detect_references: bool = typer.Option(
+        False,
+        "--detect-references",
+        help="Detect foreign keys between tables that share a key column (e.g. subject_id) and emit cr:references links. Conservative: links only when a parent table is identifiable by name.",
+    ),
     # Native mlcroissant RAI fields exposed directly as CLI flags.
     rai_data_collection: Optional[str] = typer.Option(
         None, "--rai-data-collection", help="How and where the data was gathered."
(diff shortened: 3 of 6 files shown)


rafiattrach:
@elementalcollision There are some merge conflicts. Could you look into those?

elementalcollision:
Happy to.  I’ll be able to action this tomorrow or Wednesday.

elementalcollision:
Thanks @rafiattrach — conflicts resolved in b254811. Rebased onto current `main` rather than merging, so the PR stays a single reviewable commit; the diff is unchanged in substance (+386, same four files).

**What conflicted.** The branch predated #111 and #112, which touched the same two files. All seven conflict blocks were benign:

1. **`__main__.py`** — `--jobs` (#112) and `--detect-references` landed at the same anchor (option list + the `MetadataGenerator(...)` call site). Kept both.
2. **`metadata_generator.py`** — same story for the `max_workers` / `detect_references` parameter, docstring, and assignment; the thread-pool helpers and `_apply_reference_detection` also inserted at the same spot. Kept both, upstream's first. The one true overlap: this branch added `import logging` + a module logger that `main` now already has — deduplicated to `main`'s copy.

**Verification after the rebase:**
- Full suite: **293 passed** (includes the newer determinism/thread-safety tests from #111/#112 alongside `test_references.py`). ruff clean.
- The reference-detection hook still runs post-assembly (single-threaded phase, before `_assert_unique_node_ids`), so it doesn't interact with #112's worker pool.
- e2e smoke of the flags together: `--detect-references --jobs 2` on a two-table dataset emits `samples.study_id → {"field": {"@id": "studies/study_id"}}`; the same run **without** the flag emits no `references` — opt-in default intact, committed `tests/data/output/*.jsonld` untouched.

The open design question in the description (name-based vs value-uniqueness parent selection) still stands — happy to take direction there.

rafiattrach:
@elementalcollision Thanks for the rebase, this is in good shape. Tested locally on the MIMIC-IV, eICU and OMOP demos: 14 and 52 links, all correct, output validates, default output unchanged.

**Fix before merge:** `_parent_name_variants` returns a set, so when both `study` and `studies` exist the chosen parent depends on `PYTHONHASHSEED` (I reproduced the flip across seeds). Please iterate an ordered list: exact stem, then `s`, then `ies`.

**Two small ones:**
- Drop the now-implemented references bullet from the TODO in `generate_metadata`.
- The unresolved-key messages use `logger.info` and nothing configures logging, so CLI users never see them. A warning or a line in the final summary would do.

elementalcollision:
I will get this actioned today/tomorrow.

elementalcollision:
*Reviewed by Chimera-Agent — autonomous agent, human-in-the-loop.*

Thanks @rafiattrach — all three actioned in `af02a68`, on top of a fresh rebase (`main` had moved again; #117 landed the compression/verbosity refactor, so the detection hook moved to sit right after `_disambiguate_record_sets`). Two commits now: the feature as you reviewed it, then the review response, so you can read the delta on its own.

**The headline first: your numbers are untouched.** I baked all three demos before and after, with the flag and without, and diffed the serialised documents byte for byte — **identical in all six cases**. Still 14 / 52 / 0 links and 4 / 10 / 0 unresolved keys. Everything below only changes behaviour on shapes none of those datasets contain.

---

### 1. `_parent_name_variants` — ordered list

Reproduced your flip before touching it. With `study`, `studies` and `samples` in one dataset:

```
seed=0 → studies    seed=3 → studies    seed=6 → studies
seed=1 → study      seed=4 → studies    seed=7 → study
seed=2 → study      seed=5 → study      …
```

Now an ordered list, exact stem → naive plural → grammatical plural. The same twelve seeds all pick `study`.

Two judgement calls I'd rather flag than bury:

- **I kept the `es` branch.** You named only `s` and `ies`; dropping `es` would silently stop `box_id` finding `boxes`. The `if`/`elif` collapse is set-preserving — no stem can end in both `y` and one of `s/x/z/ch/sh` — so only the *order* changed, never which names are tried.
- **Ordering is literally yours** (stem, `s`, `ies`). Putting `ies` ahead of `s` would be marginally safer against a hypothetical `studys` table. Say the word and I'll swap them; I didn't want to reinterpret the instruction.

The regression guard asserts the ordering structurally rather than sweeping seeds — a seed sweep only catches a two-way flip about half the time per seed, so it'd be a ~50% flaky test by construction.

### 2. TODO bullet

Dropped. Also retired the now-false "treats every file as an island … see the inline TODO" line in `references.py`'s module docstring, which pointed at the bullet you just had me delete.

### 3. Unresolved keys reaching the user

You offered "a warning or a line in the final summary". **The warning doesn't work here**, and it's worth saying why rather than just picking the other one: `__init__.py` attaches a `NullHandler` to the package logger, which satisfies `callHandlers`' found-counter and forecloses `logging.lastResort`. A promoted `logger.warning` is swallowed exactly like the `logger.info` was. (Same trap already bites `--field-mapping`, whose help promises "a warning prints" — separate issue, not touching it here.)

So: the summary line, following `dc29582`'s rule that the library doesn't write to a terminal. The pass records what it found on `MetadataGenerator.reference_report` (mirroring `scan_report`), and the CLI renders it — fixed-size by default, one column per line only under `--verbose`:

```
Record sets: 2
Scanned 2 file(s): 2 described, 0 not described.
Foreign keys: 0 link(s); 1 shared key(s) not linked (re-run with --verbose to name them).

# --verbose
  subject_id: shared by admissions, patients — no single record set is named after it
```

The re-run hint is folded into the line rather than added as its own `Tip:`, so it can't collide with the coverage section's. Three states stay distinguishable: `None` (pass never ran), empty (`Foreign keys: none detected.`), populated. The `len(record_sets) >= 2` half of the guard is gone so "ran and found nothing" is actually reachable.

---

### Two more wrong-parent bugs, found while pinning down yours

Both are the same class as the one you caught — the winner decided by something other than the schema — so I fixed them in the same pass. Both **narrow** what gets emitted.

**`RecordSet.name` isn't unique.** `identifiers` disambiguates `id` and leaves `name` alone, so `study.csv` and `study.tsv` become `study_csv`/`study_tsv` but stay *both named* `study`, and `setdefault` handed the parent role to whichever the filesystem yielded first:

```
study_tsv/study_id  ->  {"field": {"@id": "study_csv/study_id"}}
```

A table made a foreign-key child of itself in another format. A name two record sets share now identifies neither, and the key is reported instead. Worth noting your `PYTHONHASHSEED` fix doesn't reach this one — `study`/`studies` are two names (fixed by the ordered list), `study.csv`/`study.tsv` are the *same* name, decided by dict iteration order.

**A cycle made the whole document unreadable.** `studies(study_id, site_id)` + `sites(site_id, study_id)`:

```python
>>> ds = mlc.Dataset(out)          # document VALIDATES
>>> ds.records("studies")
NetworkXUnfeasible: Graph contains a cycle or graph changed during iteration
>>> ds.records("sites")
NetworkXUnfeasible: ...
```

Both record sets, not just the two in the cycle. Validation isn't a safety net for this. A link that would close a cycle is now refused and reported; columns are considered in name order so the surviving direction is deterministic (alphabetical, which I've said plainly in the docstring rather than dressing up as a schema judgement). Zero links dropped on your three demos — none of them takes that branch.

Smaller ones riding along: holders sorted by id so neither the links nor the reported record sets inherit `rglob` order; a column named twice in one record set no longer counts as shared with itself; the field lookup raises instead of skipping, since it resolves names copied out of descriptors built from those very `Field` objects.

---

### One thing you should know before merging

Not caused by this PR, but triggered by any `cr:references` it emits — **mlcroissant 1.1.0 mis-handles a reused `Dataset` object**. On your MIMIC-IV demo:

```python
ds = mlc.Dataset(out)
ds.records("emar")          # ok
ds.records("emar_detail")   # GenerationError → AttributeError:
                            # 'generator' object has no attribute 'columns'
```

A **fresh** `mlc.Dataset` per record set reads every one of them correctly, and with the flag off both orders work. So: the document is valid and the links are right, but a downstream consumer looping over record sets from one `Dataset` will hit this. `validate()` is not `records()` — which is also why the new tests actually read records back rather than only validating.

Happy to file it upstream with the repro. If you'd rather not ship the flag until that's resolved, I'd understand — it's your call, and it's the one thing here that might change it.

### Verification

- Full suite **591 passed** (was 568 on `main`), identical under `PYTHONHASHSEED=0` and `12345`; `pre-commit run --all-files` clean; CI green on 3.10 and 3.12.
- `tests/test_references.py` now 33 cases: variant ordering, both parent ambiguities under several input orders, a refusal still reported when another link on the same column survives, `records()` read back, the CLI in all four states, and the MIMIC-IV demo's 14 links pinned so a future heuristic change can't quietly move them.
- `docs/reference/cli.md` regenerated for the flag (`DEVELOPMENT.md` asks for it); `api.md` gains `reference_report`, since deleting the log calls makes it the only library-side channel.
- No file under `tests/data/output/` touched.

### Deliberately not in this PR

Say if you want any of these and I'll open them separately: a `references` block in the `--report` JSON (the machine-readable half — the terminal line covers your ask, but a non-TTY user still can't read this); `--verbose` naming the links it *did* make, not just the refusals; value-uniqueness parent selection (the v2 you and I have both circled — it'd also catch MIMIC's `subject_id → patients`, which name-matching can't, and would catch a non-unique parent key silently multiplying a child's rows); case-folded column matching (`Subject_ID` is currently invisible to the detector — every outcome today is a miss, never a wrong link); and `--dry-run` silently ignoring `--detect-references`.

elementalcollision:
*Reported by Chimera-Agent — autonomous agent, human-in-the-loop.*

Filed the mlcroissant caveat upstream as mlcommons/croissant#1049.

Digging into it for the report turned up that it is **wider than what I described here**, so the note above understates it:

- It is not limited to reference *chains*. A single `cr:references` link is enough for the silent half — reading the same RecordSet twice from one `mlc.Dataset` returns the rows, then **0 rows**, with no error. `mlcommons/croissant`'s own committed `datasets/1.0/titanic/metadata.json` reads 1309, then 0. So does a Hugging Face-served file (`ylecun/mnist`: 200, 0, 0 across three passes).
- The `AttributeError` I hit on `emar` → `emar_detail` is the chain case of the same root cause: `Operation.__call__` caches the target `ReadFields` output, the sequential path caches it as a raw generator, and the operation graph lives on the `Dataset`. After one read that cached object is a `GEN_CLOSED` generator.
- Reproduces identically on 1.1.0 and on `main`; present in every release since 1.0.9.

**None of this is caused by this PR, and none of it is specific to it** — it needs only a `cr:references` link, which is the feature Croissant provides for foreign keys. It does mean `--detect-references` output can meet it, so it is worth knowing. The workaround is a fresh `Dataset` per read (or `Dataset.from_metadata(ds.metadata)`, which skips re-parsing).

I offered them a candidate one-line fix; applied to `main` it fixes all three orderings and leaves their test suite unchanged. If they take it, the caveat disappears on its own.

Nothing here changes the code in this PR — flagging it only so the decision is informed.

PR #118, opened 2026-09-01
feat: add GEO SOFT metadata support

VladSam42:
Add a streaming handler for GEO `.soft` family and curated DataSet exports. Rebased onto current `main`, including #117, the 0.4.0 release in #122, and #114.

The handler describes series, sample, platform, dataset and subset attributes; submitter-defined sample characteristics; and inline tables grouped by entity kind, table title and exact column signature. `^DATABASE` is shared GEO boilerplate and is not described. Attributes and characteristics remain text. Table columns use bounded PyArrow type inference. The manifest contains schema metadata, with no attribute values or table rows and no unsupported extraction instructions.

Sample characteristics include a scalar text `geo_accession` field describing the enclosing `^SAMPLE` accession. This supplies the sample identifier alongside donor keys such as `dbgap_subject_id` or `patient id`; consumers still read the values from the source SOFT file.

Review fixes include:

- Named series tables, including GSE2034's patient clinical parameters, and all GDS dataset/subset attributes are retained. Resumed GDS dataset declarations count once; unknown entity kinds fail explicitly.
- Characteristics split on the first colon regardless of whitespace. Channel 1 keeps bare names; later channels retain prefixes. Collisions preserve distinct fields and their original locations.
- Extra empty cells after the declared table width no longer defeat type inference. Duplicate headers retain separate positional types and identifiers.
- Identifier collision suffixes consistently start at `__2`. Tests assert counts and names directly, without description-parsing regexes or external criterion references. The golden comparison resolves FileObject ids to source paths and covers reversed discovery order, so Linux and macOS compare the same metadata graph.

Validation: 721 tests passed on Python 3.10 and 721 on Python 3.12; all pinned pre-commit checks passed. Nine full public exports were checked against an independent scan of entities, attributes, characteristics and table signatures, then baked together and validated with `mlcroissant`: GSE1000, GSE2034, GSE100000, GDS10, GDS5000, GSE280315, GSE311609, GSE327347 and GSE335275. Small GSE2034/GDS10 regression extracts, provenance, updated documentation and refreshed golden output are included.


src/croissant_baker/handlers/soft_handler.py
```diff
@@ -0,0 +1,370 @@
...
+
+logger = logging.getLogger(__name__)
+
+#: GEO SOFT has no IANA registration. The ``x-`` form follows
+#: ``application/x-nifti``, already in the tree.
+ENCODING_FORMAT = "text/x-geo-soft"
+
+#: ``(entity kind, record-set suffix, noun)``. ``^DATABASE`` is absent: it is
+#: GEO boilerplate, byte-identical across a 2004 deposit and a 2026 one.
+ENTITY_RECORD_SETS = (
```
rafiattrach, line 36: Series-level tables are dropped without a word. GSE2034 has `!series_table_begin = Patient clinical parameters` with ER status, relapse and time-to-relapse per patient, which is the label table for that dataset, and the output has no trace of it. Same root cause for GDS: GDS10.soft.gz yields only `GDS10_dataset_table`, the `!dataset_*` attributes and all 12 `^SUBSET` groups are gone, and the docs don't mention GDS at all. Either emit attributes and tables for every entity kind the parser sees (DATASET, SUBSET, SERIES tables) or refuse the file with a reason. Silent partial output is the one option I'd rule out.

src/croissant_baker/handlers/soft.py
```diff
@@ -0,0 +1,388 @@
...
+_CHARACTERISTICS = re.compile(r"^characteristics(?:_ch(\w+))?$", re.IGNORECASE)
+
+#: ``!Sample_data_row_count = 22283``, likewise prefix-stripped. Every sample of
+#: every real export tested declares one, so no table body is ever counted.
+_ROW_COUNT = "data_row_count"
+
+#: What separates a characteristic's key from its value. The literal sequence,
+#: not a bare colon: a bare colon makes a key of ``stage:`` and refuses the
+#: namespaced ``efo:cell type: fibroblast``, whose key carries one of its own.
+_KEY_SEPARATOR = ": "
```
rafiattrach: `tissue:liver` (no space) lands in the fallback `characteristics_ch1` field instead of keying `tissue`. Older deposits do this, and GEOparse/GEOquery both split on the first colon regardless of the space. Same here?

src/croissant_baker/handlers/soft.py
```diff
@@ -0,0 +1,388 @@
...
+        return raw
+
+    def _name_characteristics(self) -> None:
+        """Turn the collected ``(channel, key)`` slots into field names.
+
+        A one-channel deposit keeps the bare key. Once a second channel appears
+        *every* key is prefixed, because which of two ``gender`` keys keeps the
+        bare name must not depend on which was met first.
+        """
+        prefixed = len(self._channels) > 1
```
rafiattrach: The prefix switches on for every key as soon as any sample has a ch2, so the same deposit gives `tissue` or `ch1_tissue` depending on one unrelated sample. I'd keep ch1 keys bare always and prefix only ch2 and up.

tests/test_soft_handler.py
```diff
@@ -0,0 +1,685 @@
...
+@pytest.fixture(params=[GEO_EXPORT, TABLES], ids=["tableless", "tables"])
+def any_export(request) -> Path:
+    """Both shapes, so the checks below see characteristics and tables alike."""
+    return request.param
+
+
+#: Every record set ends the same way, which is the guide's promise in code.
+_TAIL = r"\. No value is emitted\.( Partial parse: [^.]+\.)?$"
+
+_RECORD_SET_DESCRIPTIONS = [
```
rafiattrach: nit: these regexes re-parse the description prose, so any wording change breaks the suite. Asserting on the counts and names directly would be more robust. Also the "Criterion N" references in the docstrings point at a document that isn't in the repo, drop them or add it.

src/croissant_baker/handlers/utils.py
```diff
@@ -146,19 +167,93 @@ def make_record_set_ids(file_metas: list) -> list:
...
+    taken = set(bases)
+    allocated: List[Dict[str, str]] = [{} for _ in items]
+    for i in order:
+        for suffix in suffixes:
+            candidate = f"{bases[i]}_{sanitize_id(suffix)}"
+            if candidate in taken:
+                n = 2
+                while f"{candidate}__{n}" in taken:
+                    n += 1
+                candidate = f"{candidate}__{n}"
```
rafiattrach, line 232: nit: `_disambiguate_ids` numbers collisions `__1` while this and identifiers.py use `__2`. One convention would be nicer.

rafiattrach (changes requested):

src/croissant_baker/handlers/soft_handler.py
VladSam42: Added named series tables and attributes for all supported entity kinds, including DATASET and SUBSET. Resumed GDS declarations now count once; unknown kinds fail explicitly.

src/croissant_baker/handlers/soft.py
VladSam42: Characteristics now split on the first colon regardless of whitespace, so tissue:liver produces tissue.

src/croissant_baker/handlers/soft.py
VladSam42: Channel 1 keys now stay bare; channels 2+ now have the prefix. Added mixed-channel and collision tests to ensure unrelated samples cannot rename channel 1 fields.

tests/test_soft_handler.py
VladSam42: Removed description-parsing regexes and external ‘Criterion’ references.

src/croissant_baker/handlers/utils.py
VladSam42: Aligned collision suffixes to start at __2 across the shared field and record-set helpers

PR #119, opened 2026-09-02
feat: add BigTIFF and OME-TIFF header metadata

VladSam42:
Add BigTIFF (`.btf`) support and describe OME-TIFF headers in a separate `ome_images` collection. TIFF metadata is read through `tifffile`; other image formats retain Pillow. Pixel arrays are not decoded during metadata extraction.

Rebased onto `main` at `e554a5e`, including #117, the 0.4.0 release in #122, #114, and GEO SOFT support in #118.

The OME fields describe schema version, Image count, C/Z/T dimensions, dimension order, pixel type, physical X/Y spacing, and channel labels. Observed values are summarized in field descriptions. No per-file `Field.value` or unsupported header extraction instructions are emitted. Rows represent files, and Pixels fields describe the first Image in each file's XML document. Companion files and TiffData mappings are not resolved.

- X and Y retain separate unit fields. Explicit unit strings and measurements are preserved without conversion. An omitted unit remains `None`, including in older OME schemas; schema defaults are not applied. Batch ranges remain separate for each unit, with missing units reported as `unit unspecified`.
- Ordinary images retain compact extension globs with `cr:excludes` for OME files. Exclusions are resolved through compression and duplicate handling, and compression types reflect the remaining members. Literal glob characters in filenames are escaped; mixed-case image extensions remain covered.
- Image includes contain both root-level and recursive patterns. For example, `*.tif` and `**/*.tif` prevent `mlcroissant` from silently dropping a TIFF at the dataset root. The change is confined to the image handler.
- Refused OME headers fall back to `images`, with refusal counts and reasons in its description. BinaryOnly placeholders remain in `ome_images` without contributing measurements. Invalid dimension counts and non-finite or nonpositive physical sizes are omitted independently.
- XML parsing is size-bounded and requires a versioned OME namespace. Real DTD declarations are refused through the parser's declaration callback, before entity expansion. Declaration-like text inside comments, processing instructions and CDATA is accepted.

Validation: **843 tests passed on Python 3.10 and 843 on Python 3.12; all pinned pre-commit hooks passed.** Tests cover actual `mlcroissant` record reading at the dataset root and multiple directory depths, BigTIFF, uppercase extensions, annotation text, real DTDs and entity expansion. Existing coverage includes 3,000 ordinary TIFFs beside one OME file, mixed compression, linked duplicates, literal filename characters, namespaces, oversized XML, per-axis units and CLI metadata validation. Published OME classic TIFF, BigTIFF and BinaryOnly fixtures include attribution and source checksums. Two image golden outputs gain root-level include patterns; all other golden outputs are unchanged.

Reader limitations are documented and remain part of this metadata design:

- The manifest separates ordinary and OME images through includes minus `cr:excludes`. `mlcroissant` 1.1.0 validates exclusions but ignores them when reading records ([upstream issue #772](https://github.com/mlcommons/croissant/issues/772)). In mixed datasets, `records("images")` can therefore also return OME files and exceed the ordinary-image count in the description. A regression test reads actual records and verifies this behavior, so an upstream fix triggers a review of the limitation. Compact FileSets are retained rather than listing every ordinary TIFF.
- OME header fields are descriptive. `mlcroissant` 1.1.0 cannot extract them, so reading the complete `ome_images` RecordSet fails even when the metadata validates. The OME FileSet's membership is checked through the reader's own filter separately from header extraction.
- The DICOM/NIfTI header-source convention raised in review is reserved for a separate issue; draft issue text has been prepared. Those handlers and their outputs are unchanged.


src/croissant_baker/handlers/image_handler.py
```diff
@@ -239,62 +257,292 @@ def extract_metadata(self, file_path: Path, **kwargs) -> dict:
...
-            includes=includes,
-        )
+            patterns.add(f"**/*{extension}")
+
+    return mlc.FileSet(
+        id="image-files",
+        name="Image files",
+        description=f"{summary['num_images']} image files ({_formats(summary)})",
+        encoding_formats=sorted({meta["encoding_format"] for meta in file_metas}),
+        includes=sorted(patterns) + sorted(exact),
```
rafiattrach: One OME file turns the `images` FileSet from a glob into an explicit list of every plain TIFF: 3000 tiles + 1 `.ome.tif` gives 3000 entries here, and a tiled whole-slide deposit would give far more. Croissant FileSet has `excludes` and both mlcroissant and the generator already carry it, so `includes: **/*.tif` plus `excludes: <the OME paths>` keeps this O(OME files) instead of O(all files), and the OME FileSet can stay listed as it is.

src/croissant_baker/handlers/ome.py
```diff
@@ -0,0 +1,161 @@
...
+        version=namespace.rsplit("/", 1)[-1] if namespace else "",
+        image_count=len(images),
+        size_c=_int(attributes, "SizeC"),
+        size_z=_int(attributes, "SizeZ"),
+        size_t=_int(attributes, "SizeT"),
+        dimension_order=attributes.get("DimensionOrder"),
+        pixel_type=attributes.get("Type"),
+        physical_size_x=_float(attributes, "PhysicalSizeX"),
+        physical_size_y=_float(attributes, "PhysicalSizeY"),
+        physical_size_unit=attributes.get("PhysicalSizeXUnit"),
```
rafiattrach: When `PhysicalSizeXUnit` is absent this stays None and `physical_size_x` ships with no unit. The 2016-06 xsd declares `default="µm"` on that attribute, and files from before the unit attribute existed (2013-06 and older) never carry it, so those all get a bare 0.65. Default to µm when a physical size is present and the unit is not.

src/croissant_baker/handlers/image_handler.py
```diff
@@ -239,62 +257,292 @@ def extract_metadata(self, file_path: Path, **kwargs) -> dict:
...
         ]
+        if not values:
+            continue
+        fields.append(
+            mlc.Field(
+                id=f"{OME_RECORD_SET_ID}/{name}",
+                name=name,
+                description=f"{prefix} ({_observed(values)})",
+                data_types=[data_type],
+                source=mlc.Source(file_set=OME_FILE_SET_ID),
```
rafiattrach, line 525: Header fields here are `source: {fileSet}` with no extract, while DICOM and NIfTI put `fileProperty: content` on every header field. Your argument for dropping it is sound (an extract that casts an image to an integer promises nothing), but then it should be one convention for the repo, not per handler. Fine to leave as is here and open an issue to align the other two, just flagging so it's a decision rather than drift.

tests/test_image_handler.py
```diff
@@ -226,74 +168,623 @@ def test_extract_metadata_tiff(
...
+
+    assert fields["image"]["source"]["extract"] == {"fileProperty": "content"}
+    for name, field in fields.items():
+        assert "value" not in field, name
+        assert field["source"]["fileSet"] == {"@id": "ome-image-files"}
+        if name != "image":
+            assert "extract" not in field["source"], name
+
+
+def test_each_field_describes_what_the_whole_batch_holds(
```
rafiattrach, line 662: nit: this and a few neighbours assert the exact description strings, so any wording change in `_OME_FIELDS` churns them. Asserting on the observed values and field names would survive a rewording.

rafiattrach (changes requested):
Thanks! looks solid overall. Ran this against the OME sample images (2016-06 artificial series, the BinaryOnly companion pair), synthetic BigTIFF, a 2013-06 namespace, a DOCTYPE header, gzip, and 3000 plain TIFFs beside one OME file. Existing image outputs are byte-identical to main, and everything in the description holds up. One design issue with the FileSet listing, one spec gap on units, and a couple of smaller ones, inline. correct me if im wrong on the OME side.

src/croissant_baker/handlers/image_handler.py
VladSam42: Done. Ordinary images now use extension globs with cr:excludes for OME paths, including compressed files and duplicates. The 3,000-TIFF test verifies compact output.

src/croissant_baker/handlers/image_handler.py
VladSam42: Makes sense. OME header fields now retain their FileSet source without a misleading content extraction instruction. I'll create an issue to align DICOM/NIfTI accordingly and fix in the next refactor i'm planning to do before zarr

tests/test_image_handler.py
VladSam42: Updated the tests to check field names, types and observed values without depending on the explanatory wording.

src/croissant_baker/handlers/ome.py
VladSam42: fixed. now each axis is treated independently and missing units are described as None

PR #120, opened 2026-09-04
feat: HDF5 handler, with AnnData and 10x layouts (depend on #117)

VladSam42:
# An HDF5 handler

Closes `.h5`, one of the six readers G1 names. Rebased onto `main` now that #117, #118 and #119 have merged.

## What this changes

- Any `.h5`, `.h5ad` or `.hdf5` file is described by dataset path, dtype and shape. Nothing is refused.
- AnnData and 10x feature-barcode matrices get their tables instead: `obs`, `var` and `features` become record sets with named, typed columns in authored order.
- A file that only partly matches a layout falls back to the generic view rather than being claimed as one.

Only structure is read, so a 5 GB file costs what a 5 MB one with the same structure costs. `layouts.py` reads through a `Node` protocol whose methods return no values and whose attributes are asked for by name — reading them as a mapping pulls 8 MB where a named read costs 5 KiB. That module imports only the standard library, checked statically, so a Zarr reader supplying `Node` gets the same recognition.

No field carries an `extract`. mlcroissant validates one but has no HDF5 reader at any media type, so it could never execute it. Each field states its own path instead:

```text
Column 'cell_type' at /obs/cell_type in integrated.h5ad
Array /obsm/X_pca, indexed by the obs axis, in integrated.h5ad
Column 'genes' at /GRCh38/genes in filtered_gene_bc_matrices_h5.h5
```

An external link is never followed and its target never recorded — it can name any HDF5 file on the machine.

Tested on 20 real Cell Ranger and Xenium files, 953 MB: all recognised, shapes and row counts checked against the files, and describing a 297 MB matrix reads 133 KiB.

## Since @rafiattrach's review

All five comments are addressed. Verifying them turned up a sixth thing, larger than any of them.

**The legacy Cell Ranger 2 layout never fired on a real file.** It refused any container carrying a root `filetype`, on the belief that v2 wrote none. Every real one writes `filetype = matrix`. Checked against 10x's own `hgmm_1k_raw_gene_bc_matrices_h5.h5` (Cell Ranger 2.1.0) and the two files scanpy vendors: all three fell to the generic view. All three are now recognised. The same rule counted groups rather than genomes, so any extra top-level group disqualified a file as well.

**A barnyard run now gets one table pair per genome** — the nit, and the fix above is what made it reachable. Both genomes are wholly present, so both are described; the keys carry the genome only when there is more than one. scanpy and DropletUtils refuse such a file outright and Seurat returns one matrix per genome, and the two that refuse do so because their readers return exactly one matrix, which a manifest is not obliged to.

**A null dataspace no longer costs the file its description.** Exactly as you said. It is described as a scalar, and the docs name that as a loss.

**A pre-0.7 sparse `X` is one array field again.** You were right about `h5sparse_shape`; anndata 0.6.22 wrote it and 0.7.8 still reads it as backwards compat.

**An anndata 0.7 categorical is typed from its labels.** Your history is right, and 0.7.0–0.7.8 is exactly the window. One correction on the fix: the `categories` attribute cannot be tested through this reader's interface. It holds an object reference, and `Node.attr` reports a reference as absent by design — a reference names something rather than holding a value. So it matches the `__categories/<column>` sibling instead, which anndata's writer composes at that exact path. While there, `__categories` itself stopped being described as a column.

**On the fixture nit — both shapes are real, and it is pandas that decides.** anndata 0.13.3 writes `_index` and plain string columns as `string-array` datasets under pandas 2, and as `nullable-string-array` groups under pandas 3, where `StringDtype` became the default. You will have been on pandas 2. The fixture had only the second shape; it now carries both.

More broadly, the fixtures now reproduce a **dump of a real file** rather than a reading of the specs, and the dumps are in `tests/data/input/hdf5_demo/README.md`. Reading the spec was not enough: real 10x stores every string fixed-length, writes `indices` as int64, and comes from PyTables, so every object carries `CLASS`, `VERSION`, `TITLE` and `FILTERS` — `TITLE` sometimes as the very null dataspace your first comment is about. No third-party bytes are committed.

The demo dataset gains a barnyard file, so the new path is visible in the golden, and the golden comparison now normalises discovery order the way #118's does. That is not cosmetic: `discover_files` returns `rglob` order rather than a sorted one, so the old exact-text comparison passed on macOS and failed on Linux as soon as a third sample directory existed. The test runs in both orders now.

One thing not in the review: record set identifiers now come from `allocate_record_set_ids`, which #118 added for this exact problem, rather than a second local implementation of it. It grows an `include_base` for the one thing it did not return, and that drops the `.h5ad`/`.hdf5`/`.h5` entries this branch had added to `get_clean_record_name`'s hardcoded extension list.

## Size

| | Lines |
|---|---|
| `src/` | +1,309 / −3 |
| tests | +2,232 |
| fixture + golden | +863 |
| docs | +83 |
| `pyproject` + `uv.lock` | +62 |

23 files, +4,549 / −3. The suite is 950 tests.


src/croissant_baker/handlers/hdf5.py
```diff
@@ -0,0 +1,254 @@
...
+        return _normalise(self._obj.dtype)[0]
+
+    @property
+    def shape(self) -> Optional[Tuple[int, ...]]:
+        if not isinstance(self._obj, h5py.Dataset):
+            return None
+        ragged = _normalise(self._obj.dtype)[1]
+        # A variable-length dataset has one more dimension than it declares,
+        # and only the declared ones have a size.
+        return tuple(self._obj.shape) + ((-1,) if ragged else ())
```
rafiattrach: A dataset with a null dataspace (`h5py.Empty`, legal HDF5) has `shape is None`, so `tuple(...)` raises and the whole file lands in `extraction failed: 'NoneType' object is not iterable`. One such dataset should not take the other 299 with it. `()` or a skip-with-note would do?

src/croissant_baker/handlers/layouts.py
```diff
@@ -0,0 +1,677 @@
...
+            if child is not None and not child.unresolved:
+                columns.append(_array_column(f"{group}/{name}", child))
+    return columns
+
+
+def _array_shape(node: Node) -> Optional[Tuple[int, ...]]:
+    """A dense array's own shape, or a sparse group's shape attribute."""
+    if node.shape is not None:
+        return node.shape
+    declared = node.attr("shape")
```
rafiattrach: Older anndata (pre-0.7, e.g. scanpy's `pbmc3k_raw.h5ad`) stores a sparse `X` as a group with `h5sparse_format` / `h5sparse_shape` attrs and no `shape`. With `shape` absent this falls through and the obs table comes out as three columns `X/data`, `X/indices`, `X/indptr`, which are the CSR internals, not observations. Reading `h5sparse_shape` as a fallback fixes it I think

src/croissant_baker/handlers/layouts.py
```diff
@@ -0,0 +1,677 @@
...
+        return _TEXT
+    if encoding in _SPARSE:
+        return _child_type(node, "data")
+    # A group with no encoding at all is a shape no writer is known to
+    # produce, and there is nothing in it to type from. Its dtype is None,
+    # which the field builder turns into text rather than into nothing.
+    return _legacy_type(name, node.dtype, uns)
+
+
+def _legacy_type(name: str, dtype: Optional[str], uns: Optional[Node]) -> str:
```
rafiattrach: anndata 0.7.0 to 0.7.8 (2020 to 2022, a big slice of GEO h5ad deposits) wrote a categorical as an int8 dataset `obs/<col>` with a `categories` reference attr pointing into `obs/__categories/<col>`, no per-column encoding-type. That vintage comes out here as `cr:Int8`, which is the codes, exactly what the docs say never happens. Treating an integer dataset that carries a `categories` attr (or has a sibling under `__categories`) as text would cover it.

src/croissant_baker/handlers/layouts.py
```diff
@@ -0,0 +1,677 @@
...
+    Cell Ranger 2 wrote none: a container that says what it is is described by
+    what it says or not at all, never guessed at from a shape it also matches.
+    """
+    if root.attr("filetype") is not None:
+        return None
+    groups = [root.child(name) for name in root.keys()]
+    groups = [
+        g for g in groups if g is not None and not g.unresolved and g.dtype is None
+    ]
+    if len(groups) != 1:
```
rafiattrach: nit: scanpy, Seurat and DropletUtils all read a legacy 10x file with several genome groups (barnyard), one table per genome. Requiring exactly one group sends those to the generic view, which loses the fact that it is a 10x matrix at all. Rare files probably? but you can let me know since you are more familiar with this domain.

tests/hdf5_fixtures.py
```diff
@@ -0,0 +1,501 @@
...
+    _encoded(group, "dict", "0.1.0")
+    return group
+
+
+def _dataframe(parent, name: str, columns: Sequence[str], index: Sequence[str]):
+    group = parent.create_group(name)
+    _encoded(group, "dataframe")
+    group.attrs["_index"] = "_index"
+    group.attrs["column-order"] = np.asarray(columns, dtype=object)
+    _nullable(group, "_index", index, "nullable-string-array")
```
rafiattrach: nit: real anndata writes `_index` and plain string columns as `string-array` datasets, not `nullable-string-array` groups, so the committed golden isn't the shape anndata emits. The code handles the real shape (checked with anndata 0.13 output), so this only matters for what the fixture is documenting.

rafiattrach (changes requested):
Ran this on real files: 10x v3 matrices from 10x's site, scanpy's classic pbmc3k_raw.h5ad, files written by anndata 0.13 (categoricals, nullable ints and bools, layers, obsm, varm, raw, X=None), gzip and xz wrappers, plus a few edge files. The modern AnnData path and the 10x path are correct against h5py, memory stays flat, the compression numbers roughly hold. 3 things before merge, all about older or odd files, and one nit. Happy to be corrected on the anndata history.

src/croissant_baker/handlers/hdf5.py
VladSam42: Confirmed and fixed. \`h5py.Empty\` reports \`shape is None\` and \`tuple(None)\` raised out of the whole extraction, so one such dataset took every other one in the file with it.

Went with \`()\`. \`None\` was not available: the \`Node\` protocol uses it to mean "not a dataset", so a null-dataspace dataset would have been walked as a group. It is described as a scalar, and the docs now name the loss — a scalar holds one value where this holds none.

Worth adding: this is not hypothetical outside \`h5py.Empty\`. PyTables writes a null dataspace for an empty \`TITLE\`, so every dataset in a real Cell Ranger 2 file carries one as an attribute. \`write_null_dataspace\` covers the dataset case.

src/croissant_baker/handlers/layouts.py
VladSam42: Right on both counts. anndata 0.6.22 wrote \`h5sparse_format\` / \`h5sparse_shape\` (\`h5py/h5sparse.py:127-128\`) and 0.7.8 still reads them (\`_io/h5ad.py:522\`, commented "Backwards compat"). Reproduced exactly as you describe: three columns \`X/data\`, \`X/indices\`, \`X/indptr\` on the obs table.

\`_array_shape\` now falls back to \`h5sparse_shape\`. \`_sparse_dtype\` already read the \`data\` child, so the type followed once the shape was found. \`write_h5ad_h5sparse\` covers it.

src/croissant_baker/handlers/layouts.py
VladSam42: Your history is right — 0.7.0 to 0.7.8, \`__categories/<column>\` plus a \`categories\` reference attr and no per-column encoding (anndata 0.7.8 \`_io/h5ad.py:281-286\`). Reproduced: \`cr:Int8\`, which is exactly what the docs promise never happens.

One correction on the fix. The \`categories\` attr cannot be tested through this reader's interface: it holds an object reference, and \`Node.attr\` maps a reference to \`None\` by design, because a reference names something rather than holding a value. Presence is tested as \`attr(name) is not None\`, so it reads as absent. Confirmed against a real 0.7-shaped file.

So it takes your other suggestion — the \`__categories\` sibling — and types the column from that dataset. anndata composes that exact path on write, and reserves the name as a column, so it cannot collide.

Your comment also turned up a neighbour: with no \`column-order\`, \`__categories\` was itself described as a column, untyped. Excluded now.

src/croissant_baker/handlers/layouts.py
VladSam42: Doing it. Your instinct was right, and chasing it found something bigger.

I got the sources. scanpy raises \`ValueError: … contains more than one genome\` and, with \`genome=\`, reads one and discards the other; DropletUtils raises \`more than one available group (hg19, mm10)\` and does the same. Seurat is the one that returns one matrix per genome. The two that refuse do so because their readers return exactly one matrix — a constraint a manifest does not have, and both genomes are wholly present, so describing both invents nothing.

The bigger thing: **the legacy layout never fired on a real file at all.** It refused any container with a root \`filetype\`, and every real Cell Ranger 2 file writes \`filetype = matrix\` — 10x's own \`hgmm_1k_raw_gene_bc_matrices_h5.h5\` (CR 2.1.0) and both files scanpy vendors all fell to the generic view. Fixing barnyard alone would not have helped. The rule also counted groups rather than genomes, so any extra top-level group disqualified a file too.

Now: any group holding the five names qualifies, \`filetype\` refuses only when it names something other than \`matrix\`, and the keys carry the genome only when there is more than one (\`<stem>_hg19_genes\`). Verified against all five real files; three of them were unrecognised before. Not rare after all — it was every legacy file.

tests/hdf5_fixtures.py
VladSam42: Checked this one directly, and both shapes turn out to be real — it is pandas that decides, not anndata.

anndata 0.13.3 + pandas 2.3.3 → \`_index\` and plain string columns are \`string-array\` **datasets**.
anndata 0.13.3 + pandas 3.0.1 → the same columns are \`nullable-string-array\` **groups**.

pandas 3 made \`StringDtype\` the default for string columns, so anndata dispatches them to the nullable writer. You will have been on pandas 2. The fixture carried only the second shape; it now carries both — \`_index\`, \`cell_id\` and \`gene_symbol\` as datasets, plus a \`nullable_s\` group. The stale \`# anndata leaves these a string array\` comment that contradicted the code is gone.

Took the point further than the nit: the fixtures now reproduce a dump of a real file rather than a reading of the specs, with the dumps in the fixture README. That is what surfaced the fixed-length strings, the int64 \`indices\`, the PyTables attributes, and the \`filetype\` bug.

src/croissant_baker/handlers/hdf5.py
VladSam42: shape now returns the empty tuple, which this code already uses for a single value. None was taken: it means "not a dataset", so the array would have been walked as a group instead.

tests/hdf5_fixtures.py
VladSam42: It actually depends on pandas not anndata.  Under pandas 2 you get the dataset you describe; under pandas 3 you get the group, because StringDtype became the default for text columns. The hdf5_fixtures only used pandas v3 shape, and i fixed it to use both.

src/croissant_baker/handlers/layouts.py
VladSam42: i have no idea about this domain 😅. But I've added more support for legacy files. The layout refused any file with a root filetype, and every real v2 file sets filetype = matrix. So none of them were recognized, including 10x's own hgmm_1k_raw. Now filetype only disqualifies when it names something else, and every qualifying group becomes a table pair.

src/croissant_baker/handlers/layouts.py
VladSam42: _array_shape (helper function for figuring out the shape) now falls back to h5sparse_shape when there is no shape.

src/croissant_baker/handlers/layouts.py
VladSam42: used your second suggestion of matching on the __categories/<col> sibling. The attribute holds an object reference, and attr() returns None for references by design, so there is no way to tell it is there.

PR #121, opened 2026-09-07
feat: spreadsheet handler for .xlsx, .xlsm and .xls (depends on #117)

VladSam42:
# Spreadsheet handler

Handler for `.xlsx`, `.xlsm` and `.xls` files. Works on top of #117.

## What and how implemented
- Excel stops being an unclaimed file. It is how biomedical metadata travels — supplementary tables, sample manifests, variable dictionaries — and all of it landed in `no handler`.
- A sheet has no schema, so only the layout that carries almost all real data is read: one table per sheet, each its own record set. The header is the first row filling the widest row in the sheet's first hundred, which skips the title rows a supplementary file leads with; the table ends at the first blank row under it, the region a spreadsheet itself selects, so a footnote is not counted as data.
- A sheet that is not one table is named in one warning per file, and the rest of the workbook is still described. An empty sheet is skipped in silence — Excel leaves `Sheet2` behind on every new workbook.
- Fields carry no `extract`: mlcroissant has no reader for either format and Croissant cannot name a sheet, so a column reference would be a promise nothing can keep. `claims()` checks magic rather than the name, because a `.xls` that is really HTML or TSV is a common lab export.

## Size

| | Lines |
|---|---|
| `src/` | +341 / −2 |
| tests | +460 / −2 |
| docs | +19 / −1 |
| deps | +5, `uv.lock` +34 |
| test data input | +102 — two workbooks, 12 KB, one `.xls` because nothing in the stack writes that format |
| test output answers | +263 — one new file |

Net against #117's tip: **+1,224 / −5** over 17 files. The last commit reverts one
of mine that did not belong here: it sorted `discover_files` library-wide, which
reordered ten pre-existing answer files. The golden it was meant to stabilise now
uses the id-renaming and node-sorting `test_compression_matrix._normalise` has
done since PR1.


src/croissant_baker/handlers/spreadsheet_handler.py
```diff
@@ -0,0 +1,316 @@
...
+
+    An empty sheet and a sheet that is not one table are different answers: the
+    first has nothing to report, the second is what the warning is for.
+    """
+    window = list(itertools.islice(rows, _HEADER_SEARCH_ROWS))
+    width = max((_width(row) for row in window), default=0)
+    if not width:
+        return False, None
+
+    header_at = next(
```
rafiattrach, line 110: The first row with no blank cell is the header, so a header with one empty cell hands the job to the first data row, and that row's values become the column names with no signal anywhere. That shape is the default `DataFrame.to_excel()` output (blank A1 over the index), so it is common:

```
      | gene  | expr | pval        ->  columns: 0, TP53, 1.2, 0.01   (2 rows, first row gone)
0     | TP53  | 1.2  | 0.01
```

Same outcome for a two-row merged header (columns become P1, 120, 80, ...) and for a data row wider than the header. Two cheap guards would cover most of it: accept blanks in the header when the row under it fills the width and name them `column_N`; and refuse a header whose cells are mostly numbers or dates, reporting the sheet instead of describing it. Better a named skip than a wrong table.

src/croissant_baker/handlers/spreadsheet_handler.py
```diff
@@ -0,0 +1,316 @@
...
+
+        described, unreadable = [], []
+        for name, (had_content, table) in sheets:
+            if table is not None:
+                described.append({"name": name, **table})
+            elif had_content:
+                unreadable.append(name)
+
+        if unreadable:
+            logger.warning(
```
rafiattrach, line 262: This warning never reaches a user of the CLI: `__init__.py` installs a NullHandler and nothing configures logging. On your own fixture, `study.xlsx`'s `notes` sheet is dropped with no trace under default, `--verbose`, and `--report`, and the file shows as `described`. The docs promise it is named. Since #117 the pattern is to carry this on the scan entry and let the report and summary render it; a `skipped_sheets` list there, or at least the sheet names in the FileObject description, would make the promise true.

src/croissant_baker/handlers/spreadsheet_handler.py
```diff
@@ -0,0 +1,316 @@
...
+            for sheet in book.sheets()
+        ]
+    finally:
+        book.release_resources()
+
+
+def _read_with_openpyxl(source: FileSource) -> list:
+    """Read OOXML, streaming each sheet and closing the archive afterwards."""
+    with source.open() as stream:
+        book = openpyxl.load_workbook(stream, read_only=True, data_only=True)
```
rafiattrach, line 195: One of the two GEO supplementary workbooks I tried, `GSE131907_Lung_Cancer_Feature_Summary.xlsx`, fails here: openpyxl 3.1.5 raises on `docProps/custom.xml` because a custom property has no name (`StringProperty.name should be str but value is NoneType`). That part of the package carries nothing the handler needs. Catching it and retrying on an in-memory copy of the zip without `docProps/custom.xml` (and its rel) would keep real GEO files describable; worth it given the file came straight off the FTP.

src/croissant_baker/handlers/spreadsheet_handler.py
```diff
@@ -0,0 +1,316 @@
...
+
+class SpreadsheetHandler(FileTypeHandler):
+    """Handler for Excel workbooks, one record set per readable sheet."""
+
+    EXTENSIONS = (_XLS, _XLSM, _XLSX)
+    FORMAT_NAME = "Spreadsheet"
+    FORMAT_DESCRIPTION = "Per sheet: column names, inferred types, row count"
+
+    def claims(self, source: FileSource) -> bool:
+        magic = _MAGIC.get(source.suffix)
```
rafiattrach, line 228: A password-protected xlsx is an OLE2 container with the OOXML inside, so under `.xlsx` it has the `.xls` magic and lands in `no handler for file type`, which is the wrong reason for a real Excel file. Claiming OLE2 under `.xlsx`/`.xlsm` too and failing with something like `encrypted or legacy workbook` would tell the user what to do.

src/croissant_baker/handlers/spreadsheet_handler.py
```diff
@@ -0,0 +1,316 @@
...
+    header_at = next(
+        (index for index, row in enumerate(window) if _fills(row, width)), None
+    )
+    if header_at is None:
+        return True, None
+
+    columns = [str(value).strip() for value in window[header_at][:width]]
+    samples: list = [[] for _ in columns]
+    num_rows = 0
+    for row in itertools.chain(window[header_at + 1 :], rows):
```
rafiattrach, line 119: The exact row count walks every row: 200k x 20 took 19 s here, so a 2M-row export is minutes per file. CSV only counts rows under `--count-csv-rows`. Either honour that flag here as well, or cap the walk and say `at least N rows` in the description.

pyproject.toml
```diff
@@ -32,6 +32,11 @@ dependencies = [
     "pyyaml>=6.0",
     "pydicom>=2.4",
     "nibabel>=5.0",
+    "openpyxl>=3.1",
```
rafiattrach, line 35: nit, follow-up material: `python-calamine` reads xls, xlsx, xlsm and xlsb with cached values and lazy loading, and pandas ships it as an engine since 2.2. One dependency instead of two, plus `.xlsb`, which neither openpyxl nor xlrd can open.

docs/user-guide/supported-formats.md
```diff
@@ -1,25 +1,133 @@
...
+
+Each sheet that reads as one table becomes its own `cr:RecordSet`, named for the sheet, with one `cr:Field` per column. The header is the first row that fills the widest row in the sheet's first hundred, which skips the title and contact rows a supplementary table leads with. The table is what follows, up to the first blank row — the region a spreadsheet would select around a cell — so a footnote below a table is not counted into it. The row count is exact, and each column is typed by a majority of its values in the first 500 rows, except that one fraction widens a column of whole numbers.
+
+A sheet where no row fills that width — two tables side by side, say — is named in a warning and left out, and so is one with a header and nothing under it, such as a sheet holding only a note. The workbook's other sheets are still described. One warning per file names all of them. An empty sheet is skipped in silence, because Excel leaves `Sheet2` and `Sheet3` behind on every new workbook. A workbook where no sheet reads as a table is reported as a failure, with a reason, like any other unread file.
```
rafiattrach, line 115: nit: three more shapes worth naming here since they pass silently: two tables stacked with a blank row between (the wider one wins and the first is dropped), a units row under the header (`Age` / `(years)` is counted as a data row and votes in typing), and hidden or veryHidden sheets (described like any other, with no note). Also date-only columns come out `sc:DateTime`, never `sc:Date`, since both readers hand back datetime.

rafiattrach (changes requested):
Ran this on real supplementary tables from GEO (GSE164378, GSE131907), Microsoft's Financial Sample workbook, a 200k-row workbook, and a folder of shapes I see a lot in practice (title rows, units rows, merged two-row headers, a pandas export with a blank A1, stacked tables, hidden sheets, formulas without cache, gzip, xlsm, HTML and TSV impostors named .xls). The happy path is right, types are good, memory is flat, impostors are refused, xls and xlsx describe the same table the same way. Two things silently produce wrong output and one real GEO file crashes, so those first; the rest inline. Needs a rebase on main too (helpers.py and the compression matrix conflict).

src/croissant_baker/handlers/spreadsheet_handler.py
VladSam42: fixed the blank-header case, including text indexes. Missing names become column_N and the first data row stays data.

src/croissant_baker/handlers/spreadsheet_handler.py
VladSam42: Moved these onto the scan entry as structured diagnostics. The default summary counts skipped sheets, --verbose names them, and --report includes their reasons. They also appear in the FileObject description.

src/croissant_baker/handlers/spreadsheet_handler.py
VladSam42: Added the retry after confirming that the custom-properties part itself is broken. It removes that part and its package references from a temporary copy, copying the other members in bounded chunks. The original file and its checksum stay unchanged, and the recovery is reported.
Verified against GSE131907_Lung_Cancer_Feature_Summary.xlsx; it now describes successfully.

src/croissant_baker/handlers/spreadsheet_handler.py
VladSam42: fixed. OLE2 under both extensions is now claimed and reported as an encrypted or legacy workbook, with a suggestion to supply an unencrypted file with the correct extension. Added a pipeline test checking that it reaches extract_failed, not no_handler.

src/croissant_baker/handlers/spreadsheet_handler.py
VladSam42: Used the capped walk: 1,000 worksheet rows plus one row of lookahead.

docs/user-guide/supported-formats.md
VladSam42: Stacked tables are now detected across the 1,000 inspected rows and the sheet is skipped, including when the second table starts past row 100. Hidden and veryHidden sheets remain described, with their visibility stated. Documented that units, totals and contiguous footnotes count as data. Dates normally come out as sc:DateTime; ISO date strings can still produce sc:Date, so the docs distinguish those cases.

pyproject.toml
VladSam42: hm, ok. Can consider migrating later. There are some behavior changes around sheet loading, value conversion, and error handling. So, I'd keep the dependencies as is and migrate on a need-to-do basis.

PR #127, opened 2026-09-09
feat: add header-only handlers for VCF/gVCF, BCF, BAM, CRAM, SAM, FASTQ and FASTA

### LINKED ISSUE #125: genomics support
To make this even more useful from the clinical translation perspective, it could be important to add support for genomics file formats; specifically, support for BAM, VCF, gVCF, CRAM files would be very very appreciated.

renato-umeton:
Closes #125.

Adds header-only handlers for the genomic formats a sequencing delivery is made of, with no new dependencies: VCF/gVCF, BCF, BAM, CRAM, SAM, FASTQ and FASTA. Together with the existing TSV handler this covers a complete St. Jude Cloud Genomics Platform delivery (BAM or CRAM, gVCF, somatic VCF, CNV, feature counts) and the reads and reference that sit next to it. Every handler reads a header or a first record and stops; no alignment, variant or sequence record is ever read, and the read is bounded by the header, not the file.

## Variant callsets: VCF, gVCF and BCF

- `handlers/vcf_handler.py` claims on the `##fileformat=VCF` magic (so vCard `.vcf` files are refused), reads `##` and `#CHROM` header lines only, and refuses a header that does not declare the eight mandatory columns. One RecordSet per file: `CHROM`, `POS`, `ID`, `REF`, `ALT` (repeated), `QUAL`, `FILTER` (repeated), `INFO` with typed sub-fields from `##INFO`, and, when samples are present, `FORMAT` sub-fields plus one repeated `samples` field. Integer to `cr:Int64`, Float to `cr:Float64`, Flag to `sc:Boolean`, String and Character to `sc:Text`; `Number` other than 0 or 1 marks the sub-field repeated. gVCF is detected from `##GVCFBlock` or `##ALT=<ID=NON_REF`.
- `handlers/bcf_handler.py` subclasses the VCF handler. BCF is BGZF-wrapped and the compression layer does not strip `.bcf`, so the handler gunzips in place, reads the `BCF\2` magic, `l_text` (capped) and the VCF header text, and emits the same RecordSet. A `.bcf.gz` arrives with one layer already removed and is read the same way. BCF1 is claimed and refused with a reason.

## Alignments: BAM, CRAM and SAM

All three carry the same SAM text header, parsed once in `handlers/sam_header.py`. Each emits a described FileObject (sort order, reference count and assembly, read groups with platform and centre, program chain) and no RecordSet, since aligned reads are not records of a dataset schema.

- `handlers/bam_handler.py` gunzips in place, reads the magic, `l_text` (capped at 64 MiB) and the header, and stops.
- `handlers/cram_handler.py` reads the 26-byte file definition, walks the first container header (ITF8/LTF8 integers, record counter width per major version, CRC on version 3) and the FILE_HEADER block. Versions 2.x and 3.x are read; 1.x and 4.x are refused with a reason. Raw, gzip (including bare zlib, as htslib reads it), bzip2 and lzma header blocks are decoded incrementally against the size the block declares, so a block that lies about its size is refused before it is expanded; rANS-coded header blocks are refused with a reason. The reference the CRAM was encoded against is not needed, because only the header is read.
- `handlers/sam_handler.py` reads `@` lines in bounded chunks and stops at the first alignment line. Claimed on extension and header line together, since FASTQ also opens with `@`.

## Reads and references: FASTQ and FASTA

- `handlers/fastq_handler.py` reads the first record only and reports its read length. Read names are never emitted: Illumina names carry instrument, run and flowcell identifiers.
- `handlers/fasta_handler.py` reads the first line only and reports the format. Record names are never emitted: a per-sample assembly names its sample there.

## Governance default

VCF and BCF sample columns and `@RG SM` tags in BAM, CRAM and SAM form a sample manifest. By default the output carries counts only; `--genomic-sample-ids` opts in to listing identifiers. Metadata-only exports of controlled-access data therefore do not leak the manifest. FASTQ read names and FASTA record names have no opt-in and are never emitted.

## Supporting changes

- `FileSource.peek` also catches `EOFError`, `lzma.LZMAError` and `zlib.error` (named once as `sources.UNREADABLE`), so a corrupt or truncated compressed input falls through `claims` to a refusal with a reason instead of escaping dispatch; every handler catches the same set in `extract` and names the file.
- Shared helpers (`read_exactly`, `read_prefix_chunks`, `decompress_prefix`, `plural`, `MAX_HEADER_BYTES`) live in `handlers/utils.py`.
- `_file_objects_for` honours an optional `description` key from `extract`; a bake of the MIMIC-IV demo is byte-identical before and after.
- Index files (`.bai`, `.crai`, `.csi`, `.tbi`, `.fai`, `.gzi`) are reported as unclaimed; nothing opens them.
- Formats table, CLI reference and the supported-formats guide regenerated and extended with one section per format.
- The README's supported-formats table lists all seven genomic formats, and its key-features list names the sample-identifier opt-in.

## Testing

- One test module per handler (magic and extension claims, refusals with reasons, type mapping, sample opt-in, wrapped inputs, bounded reads asserted with a counting stream, memory-bounded CRAM decode asserted with `tracemalloc`, full bakes validated under mlcroissant, refusals driven through a bake and read back from the scan report), plus `SAMPLES` entries that drive the contract sweep and the compression matrix.
- `tests/data/input/genomics_htslib/` holds tiny synthetic files written by pysam 0.24.1 (htslib 1.24): SAM, BAM, CRAM 2.1, 3.0 and 3.1, VCF, BCF, FASTQ, FASTA, their index files and a CRAM 1.0 file definition. `tests/test_genomics_real_files.py` checks the handlers against them as an independent reading of the specifications: all five alignment containers describe the same header, VCF and BCF yield structurally identical RecordSets, index files are unclaimed, CRAM 1.0 is refused with its reason, and no sample identifier appears in the document without the opt-in.
- `uv run pytest -v`: 1323 passed. `uv run pre-commit run --all-files`: all hooks pass. `docs/generate.py`: no diff. Merged with `main` after the HDF5 handler landed.


rafiattrach:
@slobentanzer your review here would be greatly appreciated 🙏

renato-umeton:
Rebased on main at 0.6.0. Conflicts were in the utils imports, one docstring, the test helpers, the registry and the two docs tables, resolved by keeping both sides. No handler or test module changed beyond that. Suite green at 1455.

PR #128, opened 2026-09-09
feat: add discovery and access fields with a Bioschemas profile

### LINKED ISSUE #126: navigation of additional metadata
it would be great to discovery and access fields with Bioschemas `conformsTo`

renato-umeton:
Closes #126.

Adds five user-supplied discovery and access fields, injected post-serialisation the same way `usageInfo` and `temporalCoverage` already are. Nothing is inferred; every value is an explicit CLI input, so the traceability guarantee is unchanged.

| Flag | JSON-LD key | Shape |
|------|-------------|-------|
| `--identifier` (repeatable or comma-delimited) | `identifier` | string, or list when several are given |
| `--conditions-of-access` | `conditionsOfAccess` | string |
| `--is-accessible-for-free` / `--not-accessible-for-free` | `isAccessibleForFree` | boolean; omitted when neither flag is given |
| `--included-in-data-catalog` | `includedInDataCatalog` | URL string |
| `--profile bioschemas` (repeatable) | appends the Bioschemas Dataset profile URI to `conformsTo` | list |

## Why

Controlled-access datasets can already be described locally and published as metadata only, but the output had no field saying how access is obtained, whether it is free, or which accession (dbGaP, EGA, DOI) the data sits under. `conditionsOfAccess` and `isAccessibleForFree` are rendered by Google Dataset Search; `includedInDataCatalog` lets a release point at its catalog entry; `identifier` carries the accession. `--profile bioschemas` declares the Bioschemas Dataset profile in `conformsTo` so ELIXIR-aligned biomedical discovery services can pick the document up, which the manuscript lists as future work.

## Design notes

- Profile names resolve through one `PROFILE_CONFORMS_TO` map in `metadata_generator.py`; the CLI rejects unknown names with `typer.BadParameter` and `MetadataGenerator` raises `ValueError` for library callers, so a bad name cannot surface as a `KeyError` at serialisation.
- `conformsTo` is built once via `_resolve_conforms_to()` (Croissant 1.1 first, then declared profiles, de-duplicated in order); the RAI URI is appended afterwards by the untouched `_ensure_rai_conforms_to`, so the emitted order is Croissant, Bioschemas, RAI.
- Profile names are normalised before validation, so a padded value such as `' bioschemas'` is accepted.
- `mlcroissant.Dataset` constructs on an output that uses all five fields and round-trips them, including the two-element `conformsTo`.

## Testing

- 14 new tests in `tests/test_cli.py`: one per flag, repeat and comma forms, tri-state boolean, absence guard (no new keys when no flag is given), `conformsTo` when it starts as a string or a list and when combined with `--rai-*`, unknown profile rejected on both the CLI and library paths, padded profile accepted, and an end-to-end bake validated under mlcroissant with the Bioschemas URI asserted after the round trip.
- `uv run pytest -v`: 857 passed. `uv run pre-commit run --all-files`: all hooks pass. `docs/generate.py` regenerated `docs/reference/cli.md`; the formats and RAI tables are unchanged.

### DIFF
--- a/src/croissant_baker/__main__.py
+++ b/src/croissant_baker/__main__.py
@@ -22,6 +22,7 @@ from croissant_baker.metadata_generator import (
     MetadataGenerator,
     PROFILE_CONFORMS_TO,
     RAI_CONFORMS_TO,
+    normalize_profiles,
     serialize_datetime,
 )
 from croissant_baker import compression
@@ -315,30 +316,45 @@ def _merge_field_mapping_flags(
 _URI_SCHEME = re.compile(r"^[a-zA-Z][a-zA-Z0-9+\-.]*:")
 
 
-def _validate_uri(option_name: str, value: Optional[str]) -> None:
-    """Reject strings that don't start with an RFC 3986 URI scheme.
+def _uri_option(
+    ctx: typer.Context, param: typer.CallbackParam, value: Optional[str]
+) -> Optional[str]:
+    """Strip a URI-valued option, then reject strings with no URI scheme.
 
     Catches free text like 'see license file' but accepts http(s)://, urn:,
-    did:, mailto:, and any other valid scheme. schema.org/usageInfo accepts
+    did:, mailto:, and any other valid scheme — schema.org/usageInfo accepts
     URLs broadly, not just web URLs.
+
+    A parse-time callback rather than a check in the command body: Typer runs
+    it before ``--dry-run`` returns early, and a refusal renders as an invalid
+    option instead of being swallowed by the broad handler and reported as an
+    unexpected error. Normalising first means a blank value is absent rather
+    than a URI check nobody asked for.
     """
-    if value is None:
-        return
-    if not _URI_SCHEME.match(value):
+    value = _normalize_optional_text(value)
+    if value is not None and not _URI_SCHEME.match(value):
         raise typer.BadParameter(
-            f"{option_name} must be a URI starting with a scheme "
-            f"(e.g. https://, urn:, did:, mailto:), got {value!r}"
+            "must be a URI starting with a scheme "
+            f"(e.g. https://, urn:, did:, mailto:), got {value!r}",
+            ctx=ctx,
+            param=param,
         )
+    return value
 
 
-def _validate_profiles(values: Optional[List[str]]) -> None:
-    """Reject profile names the generator has no conformsTo URI for."""
-    unknown = sorted(set(values or []) - set(PROFILE_CONFORMS_TO))
-    if unknown:
-        raise typer.BadParameter(
-            f"--profile must be one of: {', '.join(sorted(PROFILE_CONFORMS_TO))}; "
-            f"got {', '.join(repr(name) for name in unknown)}"
-        )
+def _profile_option(
+    ctx: typer.Context, param: typer.CallbackParam, value: Optional[List[str]]
+) -> Optional[List[str]]:
+    """Normalise and check --profile while Typer parses, for the same reasons.
+
+    The rule belongs to the generator, which every caller goes through; this
+    only translates its refusal into the CLI's own error type so the message
+    is worded once.
+    """
+    try:
+        return normalize_profiles(list(value or []))
+    except ValueError as e:
+        raise typer.BadParameter(str(e), ctx=ctx, param=param)
 
 
 def _validate_iso_datetimes(option_name: str, values: Optional[List[str]]) -> None:
@@ -557,6 +573,7 @@ def main(
         None,
         "--usage-info",
         help="URI pointing to a usage or consent policy. Any RFC 3986 scheme (http(s), urn, did, mailto). Example: 'http://purl.obolibrary.org/obo/DUO_0000042' (DUO term).",
+        callback=_uri_option,
     ),
     identifier: Optional[List[str]] = typer.Option(
         None,
@@ -581,7 +598,8 @@ def main(
     profile: Optional[List[str]] = typer.Option(
         None,
         "--profile",
-        help=f"Additional profile to declare in conformsTo. One of: {', '.join(sorted(PROFILE_CONFORMS_TO))}. Declares the profile; it does not validate against it. Repeatable.",
+        help=f"Additional profile to declare in conformsTo. One of: {', '.join(sorted(PROFILE_CONFORMS_TO))}. Repeat or comma-delimit.",
+        callback=_profile_option,
     ),
     field_mappings: Optional[Path] = typer.Option(
         None,
@@ -913,11 +931,6 @@ def main(
                     err=True,
                 )
 
-        _validate_uri("--usage-info", usage_info)
-        # Normalise before validating, so the names checked here are the ones
-        # the generator is handed rather than the raw argv strings.
-        profiles = _normalize_optional_text_list(profile)
-        _validate_profiles(profiles)
         merged_field_mappings = _merge_field_mapping_flags(
             _load_field_mappings(field_mappings), field_mapping
         )
@@ -948,7 +961,7 @@ def main(
             conditions_of_access=conditions_of_access,
             is_accessible_for_free=is_accessible_for_free,
             included_in_data_catalog=included_in_data_catalog,
-            profiles=profiles,
+            profiles=profile,
             field_mappings=merged_field_mappings,
             count_csv_rows=count_csv_rows,
             max_workers=jobs or None,
--- a/src/croissant_baker/__main__.py
+++ b/src/croissant_baker/__main__.py
@@ -20,6 +20,7 @@ from rich.progress import (
 
 from croissant_baker.metadata_generator import (
     MetadataGenerator,
+    PROFILE_CONFORMS_TO,
     RAI_CONFORMS_TO,
     serialize_datetime,
 )
@@ -330,6 +331,16 @@ def _validate_uri(option_name: str, value: Optional[str]) -> None:
         )
 
 
+def _validate_profiles(values: Optional[List[str]]) -> None:
+    """Reject profile names the generator has no conformsTo URI for."""
+    unknown = sorted(set(values or []) - set(PROFILE_CONFORMS_TO))
+    if unknown:
+        raise typer.BadParameter(
+            f"--profile must be one of: {', '.join(sorted(PROFILE_CONFORMS_TO))}; "
+            f"got {', '.join(repr(name) for name in unknown)}"
+        )
+
+
 def _validate_iso_datetimes(option_name: str, values: Optional[List[str]]) -> None:
     """Validate repeated date/datetime options and raise a CLI-friendly error."""
     if not values:
@@ -546,6 +557,31 @@ def main(
         None,
         "--usage-info",
         help="URI pointing to a usage or consent policy. Any RFC 3986 scheme (http(s), urn, did, mailto). Example: 'http://purl.obolibrary.org/obo/DUO_0000042' (DUO term).",
+    ),
+    identifier: Optional[List[str]] = typer.Option(
+        None,
+        "--identifier",
+        help="Accession or persistent identifier the dataset is known by (e.g., 'phs000218.v1.p1', 'EGAS00001000255', a DOI). Repeat or comma-delimit.",
+    ),
+    conditions_of_access: Optional[str] = typer.Option(
+        None,
+        "--conditions-of-access",
+        help="How access is obtained, in free text. Example: 'Controlled access: Data Access Agreement via the Data Access Committee'.",
+    ),
+    is_accessible_for_free: Optional[bool] = typer.Option(
+        None,
+        "--is-accessible-for-free/--not-accessible-for-free",
+        help="Whether the data can be had without payment or an access agreement. Omit to leave the field out.",
+    ),
+    included_in_data_catalog: Optional[str] = typer.Option(
+        None,
+        "--included-in-data-catalog",
+        help="URL of a catalog entry listing this dataset (e.g., 'https://datacatalog.ccdi.cancer.gov/').",
+    ),
+    profile: Optional[List[str]] = typer.Option(
+        None,
+        "--profile",
+        help=f"Additional profile to declare in conformsTo. One of: {', '.join(sorted(PROFILE_CONFORMS_TO))}. Declares the profile; it does not validate against it. Repeatable.",
     ),
     field_mappings: Optional[Path] = typer.Option(
         None,
@@ -878,6 +914,7 @@ def main(
                 )
 
         _validate_uri("--usage-info", usage_info)
+        _validate_profiles(profile)
         merged_field_mappings = _merge_field_mapping_flags(
             _load_field_mappings(field_mappings), field_mapping
         )
@@ -904,6 +941,11 @@ def main(
             is_live_dataset=is_live_dataset or None,
             temporal_coverage=temporal_coverage,
             usage_info=usage_info,
+            identifier=_split_csv_list(identifier),
+            conditions_of_access=conditions_of_access,
+            is_accessible_for_free=is_accessible_for_free,
+            included_in_data_catalog=included_in_data_catalog,
+            profiles=_normalize_optional_text_list(profile),
             field_mappings=merged_field_mappings,
             count_csv_rows=count_csv_rows,
             max_workers=jobs or None,
--- a/src/croissant_baker/__main__.py
+++ b/src/croissant_baker/__main__.py
@@ -24,7 +24,7 @@ from croissant_baker.metadata_generator import (
     RAI_CONFORMS_TO,
     normalize_profiles,
     serialize_datetime,
-    url_is_iri_safe,
+    url_has_whitespace,
 )
 from croissant_baker import compression
 from croissant_baker.files import discover_files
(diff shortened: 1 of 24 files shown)


src/croissant_baker/metadata_generator.py
```diff
@@ -54,6 +54,16 @@
...
 CROISSANT_CONFORMS_TO = "http://mlcommons.org/croissant/1.1"
 RAI_CONFORMS_TO = "http://mlcommons.org/croissant/RAI/1.0"
+BIOSCHEMAS_CONFORMS_TO = "https://bioschemas.org/profiles/Dataset/1.0-RELEASE"
+
+# Profiles a document can additionally declare, by the name --profile takes.
+# A new profile is one entry here: the CLI validates against these keys and
+# the generator declares the URI alongside CROISSANT_CONFORMS_TO. Nothing else
+# in the document changes, so declaring a profile is a claim about the
+# vocabulary, not a validation against it.
+PROFILE_CONFORMS_TO = {
```
rafiattrach, line 64: Bioschemas' Dataset 1.0-RELEASE profile lists ten properties at marginality **minimum**, which in their vocabulary means must: `@context`, `@type`, `@id`, `dct:conformsTo`, `description`, `identifier`, `keywords`, `license`, `name`, `url`. Declaring the profile is what selects it; the profile page's own `dct:conformsTo` row says the versioned URL states the profile that the markup relates to.

`--profile bioschemas` on its own currently emits this:

```
conformsTo   [croissant/1.1, bioschemas/Dataset/1.0-RELEASE]
name         "MIMIC demo"
description  "Dataset containing 32 files (text/csv) ..."
identifier   <absent>
keywords     <absent>
url          <absent>
@id          <absent>
```

Three minimums missing, no warning. `@id` is the harder one: the generator never emits a top-level `@id` at all, so even a fully-flagged invocation cannot satisfy the profile today.

This matters beyond pedantry because it is machine-checked in the one place the declaration is actually consumed. FAIR-Checker reads `dct:conformsTo`, generates a SHACL shape for that profile, and reports missing mandatory triples as violations, so declaring it makes a dataset's FAIR report *worse* than declaring nothing. Google Dataset Search does not read `conformsTo`, and the Bioschemas validator makes the user pick the profile by hand, so there is no discovery upside offsetting it.

On the `@id`: it is worth doing on its own merits, and it is not something this PR broke. Parsing any current output as RDF, the dataset comes back as a blank node:

    Nc22e15e3f52e4ae2a763a336291371d3  ->  sc:name "MIMIC-IV Clinical Database Demo"

So every file the tool has produced so far describes an anonymous dataset that nothing external can link to or cite. Same on main, so it long predates this branch. Since you are already in this part of the code, closing it here would be a nice win well beyond Bioschemas.

Suggestion: make the declaration conditional. Emit a top-level `@id` (the dataset `url` is the natural value), and have `--profile bioschemas` refuse, or at minimum warn loudly, when `identifier`, `keywords`, `url`, `license` or `description` are missing. That turns a decorative flag into one a user can rely on.

src/croissant_baker/metadata_generator.py
```diff
@@ -548,6 +589,18 @@ def generate_metadata(self, progress_callback=None) -> dict:
...
+            # own single-element lists; several stay a list.
+            result["identifier"] = (
+                self.identifier[0] if len(self.identifier) == 1 else self.identifier
+            )
+        if self.conditions_of_access is not None:
+            result["conditionsOfAccess"] = self.conditions_of_access
+        if self.is_accessible_for_free is not None:
+            result["isAccessibleForFree"] = self.is_accessible_for_free
+        if self.included_in_data_catalog is not None:
+            result["includedInDataCatalog"] = self.included_in_data_catalog
```
rafiattrach, line 603: schema.org gives `includedInDataCatalog` exactly one entry in `rangeIncludes`, and it is `DataCatalog`. Not Text, not URL. (Checked against `schemaorg-current-https.jsonld`. Contrast `identifier`, which does list Text and URL, and `conditionsOfAccess`, which lists Text: both of those are correct as bare strings here.)

So this emits a plain literal where a consumer expects a node:

```json
"includedInDataCatalog": "https://physionet.org/content/mimic-iv-demo/"
```

Under the emitted `"@vocab": "https://schema.org/"` that expands to a language-tagged string, not a catalog link, which undercuts the discovery purpose of the flag.

The repo already does this correctly one field over: `--publisher PhysioNet` becomes `{"@type": "sc:Organization", "name": "PhysioNet"}`. Same treatment here:

```json
"includedInDataCatalog": {"@type": "sc:DataCatalog", "url": "https://physionet.org/content/mimic-iv-demo/"}
```

I round-tripped that nested shape through mlcroissant and it survives untouched.

Separately, the help text calls it a URL but nothing checks it: `--included-in-data-catalog "not a url"` exits 0 and writes the string straight through. `--usage-info` is the only other option whose help says URI and it runs `_validate_uri`. Worth the same call, or drop URL from the help.

src/croissant_baker/metadata_generator.py
```diff
@@ -548,6 +589,18 @@ def generate_metadata(self, progress_callback=None) -> dict:
             result["temporalCoverage"] = self.temporal_coverage
         if self.usage_info is not None:
             result["usageInfo"] = self.usage_info
+        if self.identifier:
+            # One accession reads as a string, the way mlcroissant flattens its
+            # own single-element lists; several stay a list.
+            result["identifier"] = (
```
rafiattrach, line 595: `tests/test_mlcroissant_api_gaps.py` is the tripwire for exactly this block. It lists the fields injected post-hoc because mlcroissant has no native parameter, and fails once upstream grows one, so the workaround gets deleted rather than quietly outliving its reason. It currently names `alternate_name`, `is_live_dataset`, `temporal_coverage`, `usage_info`.

These four new injects are not added to it. I confirmed with `inspect.signature(mlc.Metadata.__init__)` that none of `identifier`, `conditions_of_access`, `is_accessible_for_free`, `included_in_data_catalog` exists in mlcroissant 1.1.0, so all four qualify.

One wrinkle worth deciding explicitly rather than leaving implicit: that file's docstring says "fields the Croissant 1.1 spec defines", and these four are schema.org properties reached through `@vocab` rather than spec-table entries. Either widen the docstring to cover both kinds or add a second set for schema.org passthroughs. Either way the injects should not be invisible to the mechanism built to catch them.

src/croissant_baker/metadata_generator.py
```diff
@@ -253,6 +283,17 @@ def __init__(
         self.is_live_dataset = is_live_dataset
         self.temporal_coverage = temporal_coverage
         self.usage_info = usage_info
+        self.identifier = identifier
+        self.conditions_of_access = conditions_of_access
+        self.is_accessible_for_free = is_accessible_for_free
+        self.included_in_data_catalog = included_in_data_catalog
+        unknown = sorted(set(profiles or []) - set(PROFILE_CONFORMS_TO))
```
rafiattrach, line 290: Profile names are validated twice and the two checks disagree. `__main__.py` normalises first and then validates, so this branch is unreachable from the CLI and fires only for library callers, who then get a different contract:

```
$ croissant-baker ... --profile " bioschemas "
accepted (and pinned by a test)

>>> MetadataGenerator(path, profiles=[" bioschemas "])
ValueError: Unknown profile(s)  bioschemas ; known profiles: bioschemas
```

The two messages differ in wording as well. And a caller who passes a bare string rather than a list gets this, because `set()` iterates characters:

```
>>> MetadataGenerator(path, profiles="bioschemas")
ValueError: Unknown profile(s) a, b, c, e, h, i, m, o, s; known profiles: bioschemas
```

Simplest fix is a single owner: strip inside the generator, delete `_validate_profiles`, and let the CLI surface the generator's error. Storing the normalised list in `self.profiles` would also close the `PROFILE_CONFORMS_TO[profile]` lookup in `_resolve_conforms_to`, which is safe today only because `__init__` happened to validate first.

src/croissant_baker/__main__.py
```diff
@@ -878,6 +914,10 @@ def main(
                 )
 
         _validate_uri("--usage-info", usage_info)
+        # Normalise before validating, so the names checked here are the ones
+        # the generator is handed rather than the raw argv strings.
+        profiles = _normalize_optional_text_list(profile)
+        _validate_profiles(profiles)
```
rafiattrach, line 920: This sits after the `--dry-run` early return, so a bad profile is not caught in dry-run at all:

```
$ croissant-baker -i data --dry-run --profile biocroissant
... scans normally, exit 0, no complaint
```

Dry-run is where someone checks their flags before committing to a run, so it is the one place the check earns its keep.

Related: the `typer.BadParameter` raised here is swallowed by the broad `except Exception` below, so the user sees `Unexpected error: --profile must be one of: bioschemas; got 'BIOSCHEMAS'`. It is not unexpected, it is their typo.

A `callback=` on the `typer.Option` fixes both at once: Typer runs it during parsing, which covers dry-run, and renders it as a proper `Invalid value for '--profile'`. `--usage-info` has the same "Unexpected error" wart today, so this is a chance to set the better pattern rather than a regression you introduced.

src/croissant_baker/metadata_generator.py
```diff
@@ -548,6 +589,18 @@ def generate_metadata(self, progress_callback=None) -> dict:
...
         if self.usage_info is not None:
             result["usageInfo"] = self.usage_info
+        if self.identifier:
+            # One accession reads as a string, the way mlcroissant flattens its
+            # own single-element lists; several stay a list.
+            result["identifier"] = (
+                self.identifier[0] if len(self.identifier) == 1 else self.identifier
+            )
+        if self.conditions_of_access is not None:
+            result["conditionsOfAccess"] = self.conditions_of_access
```
rafiattrach, line 599: The guard is `is not None` and the CLI passes the raw argv value with no normalisation, so an empty or whitespace-only flag becomes an empty property rather than an absent one:

```
--conditions-of-access ""      ->  "conditionsOfAccess": ""
--conditions-of-access "   "   ->  "conditionsOfAccess": "   "
```

Same for `--included-in-data-catalog`. `_normalize_optional_text` already exists for this and every RAI free-text flag uses it; two calls in `__main__.py` and both become absent instead.

tests/test_cli.py
```diff
@@ -938,3 +943,207 @@ def test_baked_output_round_trips_through_mlcroissant(
...
+    assert not {
+        "identifier",
+        "conditionsOfAccess",
+        "isAccessibleForFree",
+        "includedInDataCatalog",
+    } & set(metadata)
+    assert metadata["conformsTo"] == "http://mlcommons.org/croissant/1.1"
+
+
+def test_all_discovery_fields_construct_under_mlcroissant(
```
rafiattrach, line 1124: All thirteen new CLI tests go through the `cli` helper, which hardcodes `--no-validate`, this one included. So the round-trip is checked through `mlc.Dataset` directly, but the default save path, the one every real user hits, is never exercised with the new fields.

I ran it by hand and it passes today: `--validate` with a three-element `conformsTo` (Croissant + Bioschemas + RAI) validates clean. Worth pinning at least this test without `--no-validate` so a later change to the validate path cannot break it silently.

src/croissant_baker/__main__.py
```diff
@@ -878,6 +914,10 @@ def main(
                 )
 
         _validate_uri("--usage-info", usage_info)
+        # Normalise before validating, so the names checked here are the ones
+        # the generator is handed rather than the raw argv strings.
+        profiles = _normalize_optional_text_list(profile)
```
rafiattrach, line 919: nit: two small inconsistencies between the two list options added here.

`--profile` uses `_normalize_optional_text_list` (no comma splitting) while `--identifier`, added in the same PR, uses `_split_csv_list` (comma splitting), as do `--keywords` and `--same-as`. So `--profile "bioschemas,bioschemas"` is rejected as an unknown profile. Only one profile exists today so nobody can hit it yet, but the two habits will collide the moment a second one lands.

And `--identifier` does not deduplicate, which is consistent with `--keywords`, except that here it also flips the JSON type because of the single-value special case:

```
--identifier a                 ->  "identifier": "a"
--identifier a --identifier a  ->  "identifier": ["a", "a"]
```

A `dict.fromkeys` pass, the same one `_resolve_conforms_to` already uses for profiles, keeps the shape predictable.

tests/test_cli.py
```diff
@@ -938,3 +943,207 @@ def test_baked_output_round_trips_through_mlcroissant(
...
+    """An unrecognised profile name fails loudly and names what is accepted."""
+    output = tmp_path / "output.jsonld"
+
+    result = cli(csv_dataset, output, "--profile", "biocroissant")
+
+    assert result.exit_code != 0
+    assert "bioschemas" in result.output
+
+
+def test_unknown_profile_is_rejected_by_the_generator(tmp_path: Path) -> None:
```
rafiattrach, line 1086: nit: three small things in this block.

This matches on `"bioschemas"`, but that string also appears in the *known profiles* half of the message, so the test passes even if the rejected name were dropped from the error entirely. Matching on `biocroissant` would test what it is named for.

`test_padded_profile_name_is_accepted` passes `" bioschemas"`, leading whitespace only, despite the name.

The bioschemas URI is hardcoded in several assertions here while `BIOSCHEMAS_CONFORMS_TO` is imported at the top of the file and used once, further down.

rafiattrach (changes requested):
Ran this on the MIMIC-IV clinical demo from PhysioNet with real accessions (a DOI, a dbGaP phs number, an EGA study id), then read every output back through mlcroissant and expanded it to RDF to check the new properties actually survive.

They do. `conformsTo` as a list is explicitly legal in Croissant 1.1 (cardinality MANY, and the spec says as much in so many words), mlcroissant still detects 1.1 regardless of list order, all four schema.org properties round-trip unchanged, the tri-state boolean behaves exactly as documented, the profile list composes correctly with the RAI append, and output with none of the new flags is byte-identical to main. 857 tests pass.

Two things I'd change before merge: the Bioschemas declaration as it stands is a claim the tool cannot back, and `includedInDataCatalog` is the wrong shape for what schema.org expects there. The rest is smaller, and the last three are marked nit.

src/croissant_baker/metadata_generator.py
renato-umeton: Done. The Dataset now carries a top-level `@id` set to `url` whenever `--url` is given. It is injected after serialisation: mlcroissant 1.1.0 emits no `@id` even from its native `id` parameter, and a tripwire in `test_mlcroissant_api_gaps.py` fails the day it starts to, so the inject gets deleted rather than outliving its reason. A url carrying whitespace cannot be an IRI, so it gets no `@id` and the CLI warns, suggesting percent-encoding; `@id` is one of the checked minimums, so `--profile bioschemas` refuses in that case too.

`--profile bioschemas` now refuses when the assembled document lacks any of identifier, keywords, url, license or description, naming the missing ones:

    Error: Profile 'bioschemas' requires fields this document does not carry: identifier, keywords, url.

The check reads the finished document, so it runs on a bake and not on `--dry-run`, which assembles none (the profile name check does cover dry-run, see below). With the fields supplied, the output carries all ten minimums and validates under mlcroissant. The `PROFILE_CONFORMS_TO` comment and the `--profile` help no longer say the profile is declared without checks.

src/croissant_baker/metadata_generator.py
renato-umeton: Emitted as `{"@type": "sc:DataCatalog", "url": ...}` now, and the option runs the same URI callback as `--usage-info`, so `--included-in-data-catalog "not a url"` is rejected while parsing.

src/croissant_baker/metadata_generator.py
renato-umeton: The four are in `_METADATA_GAPS_AS_OF_MLC_1_1_0`. One set, with the comment widened to say it holds both spec-table fields and schema.org properties reached through `@vocab`, since the action when mlcroissant grows a parameter is the same for both. A separate test covers `@id`: it asserts `mlc.Metadata(id=...).to_json()` still emits no top-level `@id`.

src/croissant_baker/metadata_generator.py
renato-umeton: `normalize_profiles` in `metadata_generator.py` is the one owner now: it strips, splits comma lists, drops empties, deduplicates in order, validates, and raises one `ValueError`. A bare string is read as one name. `self.profiles` stores the normalised list, so the lookup in `_resolve_conforms_to` is safe by construction. `_validate_profiles` is gone from `__main__.py`, and the CLI prints the generator's message.

src/croissant_baker/__main__.py
renato-umeton: `--profile` and `--usage-info` validate through Typer callbacks now, so a bad value is caught while parsing, `--dry-run` included, and renders as

    Invalid value for '--profile': Unknown profile(s): 'biocroissant'. Known profiles: bioschemas

with exit code 2. The callback calls the generator's normaliser and turns its `ValueError` into `BadParameter`, so the text has one source. `--included-in-data-catalog` uses the same URI callback.

src/croissant_baker/metadata_generator.py
renato-umeton: Both go through `_normalize_optional_text` now: an empty or whitespace value leaves the key absent.

tests/test_cli.py
renato-umeton: The `cli` helper takes `validate=True`, and `test_all_discovery_fields_construct_under_mlcroissant` uses it, asserting on the line only a validating bake prints.

src/croissant_baker/__main__.py
renato-umeton: `--profile` accepts comma lists through the normaliser, and `identifier` is deduplicated in order in the generator, so `--identifier a --identifier a` emits `"a"` and `--identifier a,b --identifier a` emits `["a", "b"]`.

tests/test_cli.py
renato-umeton: All three: the rejection test asserts on `biocroissant`, the padded test passes `" bioschemas "`, and the conformsTo assertions use `CROISSANT_CONFORMS_TO`, `BIOSCHEMAS_CONFORMS_TO` and `RAI_CONFORMS_TO`.

PR #130, opened 2026-09-09
feat: add a local MCP server and an Agent Skill

### LINKED ISSUE #129: MCPfy and SKILLify the whole thing
this would make the thing easier to integrate with LLMs

renato-umeton:
Closes #129.

Two halves, both local-first and both leaving the deterministic structural pipeline untouched.

## MCP: `croissant-baker mcp`

- A stdio Model Context Protocol server exposing three tools over the existing `MetadataGenerator`: `dry_run(input_dir, include?, exclude?)` returns a summary (`total`, `would_process`, `unclaimed`, `by_reason`) plus the per-file list with reasons; `bake(...)` accepts the semantic fields a human would type on the CLI, runs generation, validates before writing, and returns the output path and report; `validate(path)` constructs `mlcroissant.Dataset` and returns success or the error text.
- stdio only: no HTTP listener, no outbound requests, no fetch or search tool. An agent can only supply what a human could type, so every value in the output remains traceable to file bytes or an explicit input.
- The `mcp` SDK lives in an optional dependency group (`uv sync --group mcp`); runtime dependencies are unchanged and the subcommand fails with an install hint when the group is absent. Imports are lazy, so `croissant-baker --help` and a normal bake never load `mcp`.
- CI installs the group so the MCP tests run on both matrix jobs.
- The `mcp` 2.x SDK renamed `FastMCP` to `MCPServer`; the code uses the latter.

## Skill: `croissant-baker`

- An Agent Skill following the specification at https://agentskills.io/specification: `SKILL.md` with `name`, `description`, `license`, `compatibility`, `metadata` and `allowed-tools` frontmatter, a procedural body (dry run, resolve refusals, gather semantic fields from the user, bake, validate, iterate), a Gotchas section, and one asset (`assets/rai-template.yaml`).
- The canonical copy ships inside the package (`croissant_baker/skills/croissant-baker/`) so `pip install croissant-baker` users receive it; `.agents/skills/croissant-baker` is a symlink to it for repository-level discovery by clients that scan that path.
- The MCP server exposes the same `SKILL.md` as the resource `croissant-baker://skill`, so an MCP-only client can fetch the instructions without the repository.
- No bundled scripts: the CLI already is the non-interactive script the specification asks for (`--dry-run`, `--report` JSON, `validate`, documented exit codes).
- Validated with the reference validator `skills-ref validate` and guarded by `tests/test_skill.py` (frontmatter constraints, name matches directory, body length, referenced paths exist, symlink resolves, packaged copy equals the repository copy).

## Testing

- `tests/test_mcp_server.py` (tool functions called directly, no transport; `build_server()` lists exactly the three tools and the skill resource; `dry_run` counters match per-file outcomes), `tests/test_skill.py`, and a CLI test that `croissant-baker mcp --help` mentions stdio.
- `uv run pytest -v` with the `mcp` group: 860 passed; without it the MCP and skill-resource tests skip and the rest is unchanged. `skills-ref validate`: valid skill, no warnings.
- `uv run pre-commit run --all-files`: all hooks pass.

## Design notes

- `_dry_run_entries` and `_parse_creators` were extracted from `main()` so the tools share the CLI's exact behaviour; `--dry-run` and `--creator` parsing are unchanged.
- `uv.lock` was relocked with a current `uv` so the lock revision matches `main`; the diff is confined to the optional group plus the pins `mcp` itself forces.


pyproject.toml
```diff
@@ -46,6 +46,9 @@ dev = [
     "pre-commit>=4.5.1",
     { include-group = "test" },
 ]
+# Optional: the local stdio MCP server (`croissant-baker mcp`).
+# Install with: uv sync --group mcp
+mcp = ["mcp>=2.2"]
```
rafiattrach: This is under `[dependency-groups]`, and the comment four lines above says what that means: PEP 735 groups are not propagated to downstream consumers. So the extra does not exist for anyone who installs the package.

From the wheel this branch builds:

```
Provides-Extra:               NONE
Requires-Dist mentioning mcp: NONE

$ uv pip install "croissant_baker-0.4.0-py3-none-any.whl[mcp]"
Audited 1 package
$ python -c "import mcp"
ModuleNotFoundError
```

It fails silently, which is worse than failing loudly. `pip install croissant-baker[mcp]` looks like it worked and the server still will not start.

Moving it to `[project.optional-dependencies]` gives a real `croissant-baker[mcp]` extra in the wheel metadata, and the dependency group can stay as an alias for contributors. Both `SKILL.md` and `DEVELOPMENT.md` advertise this to end users, so the install path needs to work for them, not only inside a source checkout.

Related, and part of the same gap: `grep -i "mcp\|skill\|agent" README.md` returns nothing. The README is the only page a PyPI user reads, and right now the feature appears in `DEVELOPMENT.md` (a contributor document) and in the generated `--help` dump, neither of which says how to install it or how to point a client at it. A short README section would close both halves.

src/croissant_baker/mcp_server.py
```diff
@@ -0,0 +1,194 @@
...
+    source checkout. Nothing is read until a client asks for the resource.
+    """
+    return (
+        importlib.resources.files("croissant_baker")
+        .joinpath("skills", "croissant-baker", "SKILL.md")
+        .read_text(encoding="utf-8")
+    )
+
+
+def dry_run(
```
rafiattrach: Every failure in these tools reaches the agent as a bare string with all the useful text stripped out. Driven over real stdio:

```
dry_run(input_dir="/does/not/exist")  ->  Error executing tool dry_run
dry_run(input_dir="pyproject.toml")   ->  Error executing tool dry_run
bake(input_dir="/does/not/exist", …)  ->  Error executing tool bake
```

The real messages underneath are good and would tell an agent exactly what to do (`Directory not found: /does/not/exist is not a directory`, `Dataset path pyproject.toml is not a directory`), and none of it survives. The SDK draws the line explicitly: `ToolError` is the failure you anticipated and its message goes to the model, anything else is treated as a crash and the model sees only `Error executing tool <name>`. There is no `ToolError` anywhere in this file.

An agent told only "Error executing tool bake" cannot tell a typo'd path from a permission problem from a validation failure, so it retries blindly or gives up. Wrapping the anticipated failures is a few lines, and `validate` already gets this right by returning `{"valid": False, "error": ...}` instead of raising, so the other two just need to match it.

Worth noting `bake`'s docstring advertises `Raises: ValueError: If the document fails mlcroissant validation`, which is an error contract the agent provably cannot observe.

src/croissant_baker/mcp_server.py
```diff
@@ -0,0 +1,194 @@
...
+            case nothing is written.
+    """
+    generator = MetadataGenerator(
+        dataset_path=input_dir,
+        name=name,
+        description=description,
+        url=url,
+        license=license,
+        citation=citation,
+        creators=_parse_creators(creators) or None,
```
rafiattrach: An empty `creators` list produces a document with a placeholder creator in it, where the CLI refuses outright:

```
CLI:  croissant-baker -i data --name T
      Error: At least one '--creator' option is required to comply with the
      Croissant specification.

MCP:  bake(input_dir="data", name="T", creators=[], …)
      succeeds, and writes:
      "creator": {"@type": "sc:Person", "name": "Dataset Creator",
                  "email": "creator@example.com"}
```

`creators` being required in the schema does not help, since `[]` satisfies it. The result is a published file naming a person who does not exist, produced by the path with no human watching it, which is the wrong way round.

This is also the exact drift the extraction in `__main__.py` was meant to prevent, so it is worth fixing at the same place the CLI does its check rather than in both.

src/croissant_baker/mcp_server.py
```diff
@@ -0,0 +1,194 @@
...
+        dependency stays optional.
+
+    Raises:
+        ImportError: If the optional ``mcp`` dependency group is not installed.
+    """
+    from mcp.server.mcpserver import MCPServer
+
+    server = MCPServer(SERVER_NAME)
+    for tool in (dry_run, bake, validate):
+        server.add_tool(tool)
```
rafiattrach: No tool carries annotations, so a client cannot tell the two read-only tools from the one that writes to disk. All three come back with `annotations = None`.

`dry_run` and `validate` read nothing and change nothing, and `dry_run`'s docstring makes a point of saying it does not even read file contents. `bake` writes a file, and it will create directories to do it: I called it with an output under a path three levels deep that did not exist and it made all three.

`readOnlyHint=True` on the first two and `destructiveHint=True` on `bake` is what lets a client auto-approve the safe ones and prompt on the third. Without them a client has to treat all three the same, so it either gates everything, which is friction, or approves everything, which means the writing one never prompts.

nit: while you are in here, `bake` has no `date_published` parameter, but `SKILL.md` tells the agent "Give `--url` and `--date-published`". Over MCP it cannot, so every document the server produces carries the `datePublished` recommendation warning with no way to comply. One more parameter closes it.

src/croissant_baker/skills/croissant-baker/SKILL.md
```diff
@@ -0,0 +1,229 @@
...
+  --url "https://physionet.org/content/mimic-iv-demo/2.2/" \
+  --license "https://opendatacommons.org/licenses/odbl/1-0/" \
+  --date-published "2023-01-06" \
+  --rai-data-biases "Single-site cohort from a US academic medical centre" \
+  --rai-data-limitations "Demo subset limited to 100 patients" \
+  --output mimic-iv-demo-croissant.jsonld
+```
+
+`--creator` is `Name[,Email[,URL]]` and repeats, once per creator. Prefer a
+license URL over an SPDX string. Give `--url` and `--date-published`: the
```
rafiattrach: nit: two test gaps worth closing, both cheap, both guarding claims made elsewhere in this PR.

Nothing compares the CLI and MCP outputs. I checked they match today, but `tests/test_mcp_server.py` calls the functions directly and never diffs against the CLI, so the parity that justifies the whole refactor is unpinned. RAI injection and `_ensure_rai_conforms_to` sit in the CLI's bake path and are exactly the kind of step that gets added to one side later. Baking one fixture both ways and asserting the files are byte-identical is about fifteen lines.

Nothing checks the flags in this file against the real CLI either. All 27 of them are correct today, I diffed them against `--help`, but a rename would leave this file quietly telling agents to pass a flag that no longer exists, with no CI signal. Pulling `--[a-z0-9-]+` out of the body and asserting each one is a registered option would cover it.

Smaller, same file: `test_the_installed_package_carries_the_skill` cannot fail. Under the editable install `uv sync` creates, `importlib.resources.files("croissant_baker")` resolves back to `src/croissant_baker`, so it compares the file to itself and passes whether or not the wheel ships anything. And `assets/rai-template.yaml` is only asserted to exist, never loaded, even though its own header warns that a misspelled key is dropped in silence.

renato-umeton:
Rebased on main at 0.6.0. Only uv.lock conflicted; it was relocked with the mcp group on top of main's lock. Suite green at 1100.

pyproject.toml
```diff
@@ -50,6 +50,9 @@ dev = [
     "pre-commit>=4.5.1",
     { include-group = "test" },
```
rafiattrach: This is under `[dependency-groups]`, and the comment four lines above says what that means: PEP 735 groups are not propagated to downstream consumers. So the extra does not exist for anyone who installs the package.

From the wheel this branch builds:

```
Provides-Extra:               NONE
Requires-Dist mentioning mcp: NONE

$ uv pip install "croissant_baker-0.4.0-py3-none-any.whl[mcp]"
Audited 1 package
$ python -c "import mcp"
ModuleNotFoundError
```

It fails silently, which is worse than failing loudly. `pip install croissant-baker[mcp]` looks like it worked and the server still will not start.

Moving it to `[project.optional-dependencies]` gives a real `croissant-baker[mcp]` extra in the wheel metadata, and the dependency group can stay as an alias for contributors. Both `SKILL.md` and `DEVELOPMENT.md` advertise this to end users, so the install path needs to work for them, not only inside a source checkout.

Related, and part of the same gap: `grep -i "mcp\|skill\|agent" README.md` returns nothing. The README is the only page a PyPI user reads, and right now the feature appears in `DEVELOPMENT.md` (a contributor document) and in the generated `--help` dump, neither of which says how to install it or how to point a client at it. A short README section would close both halves.

src/croissant_baker/mcp_server.py
rafiattrach: Every failure in these tools reaches the agent as a bare string with all the useful text stripped out. Driven over real stdio:

```
dry_run(input_dir="/does/not/exist")  ->  Error executing tool dry_run
dry_run(input_dir="pyproject.toml")   ->  Error executing tool dry_run
bake(input_dir="/does/not/exist", …)  ->  Error executing tool bake
```

The real messages underneath are good and would tell an agent exactly what to do (`Directory not found: /does/not/exist is not a directory`, `Dataset path pyproject.toml is not a directory`), and none of it survives. The SDK draws the line explicitly: `ToolError` is the failure you anticipated and its message goes to the model, anything else is treated as a crash and the model sees only `Error executing tool <name>`. There is no `ToolError` anywhere in this file.

An agent told only "Error executing tool bake" cannot tell a typo'd path from a permission problem from a validation failure, so it retries blindly or gives up. Wrapping the anticipated failures is a few lines, and `validate` already gets this right by returning `{"valid": False, "error": ...}` instead of raising, so the other two just need to match it.

Worth noting `bake`'s docstring advertises `Raises: ValueError: If the document fails mlcroissant validation`, which is an error contract the agent provably cannot observe.

src/croissant_baker/mcp_server.py
rafiattrach: An empty `creators` list produces a document with a placeholder creator in it, where the CLI refuses outright:

```
CLI:  croissant-baker -i data --name T
      Error: At least one '--creator' option is required to comply with the
      Croissant specification.

MCP:  bake(input_dir="data", name="T", creators=[], …)
      succeeds, and writes:
      "creator": {"@type": "sc:Person", "name": "Dataset Creator",
                  "email": "creator@example.com"}
```

`creators` being required in the schema does not help, since `[]` satisfies it. The result is a published file naming a person who does not exist, produced by the path with no human watching it, which is the wrong way round.

This is also the exact drift the extraction in `__main__.py` was meant to prevent, so it is worth fixing at the same place the CLI does its check rather than in both.

src/croissant_baker/mcp_server.py
rafiattrach: No tool carries annotations, so a client cannot tell the two read-only tools from the one that writes to disk. All three come back with `annotations = None`.

`dry_run` and `validate` read nothing and change nothing, and `dry_run`'s docstring makes a point of saying it does not even read file contents. `bake` writes a file, and it will create directories to do it: I called it with an output under a path three levels deep that did not exist and it made all three.

`readOnlyHint=True` on the first two and `destructiveHint=True` on `bake` is what lets a client auto-approve the safe ones and prompt on the third. Without them a client has to treat all three the same, so it either gates everything, which is friction, or approves everything, which means the writing one never prompts.

nit: while you are in here, `bake` has no `date_published` parameter, but `SKILL.md` tells the agent "Give `--url` and `--date-published`". Over MCP it cannot, so every document the server produces carries the `datePublished` recommendation warning with no way to comply. One more parameter closes it.

src/croissant_baker/skills/croissant-baker/SKILL.md
rafiattrach: nit: two test gaps worth closing, both cheap, both guarding claims made elsewhere in this PR.

Nothing compares the CLI and MCP outputs. I checked they match today, but `tests/test_mcp_server.py` calls the functions directly and never diffs against the CLI, so the parity that justifies the whole refactor is unpinned. RAI injection and `_ensure_rai_conforms_to` sit in the CLI's bake path and are exactly the kind of step that gets added to one side later. Baking one fixture both ways and asserting the files are byte-identical is about fifteen lines.

Nothing checks the flags in this file against the real CLI either. All 27 of them are correct today, I diffed them against `--help`, but a rename would leave this file quietly telling agents to pass a flag that no longer exists, with no CI signal. Pulling `--[a-z0-9-]+` out of the body and asserting each one is a registered option would cover it.

Smaller, same file: `test_the_installed_package_carries_the_skill` cannot fail. Under the editable install `uv sync` creates, `importlib.resources.files("croissant_baker")` resolves back to `src/croissant_baker`, so it compares the file to itself and passes whether or not the wheel ships anything. And `assets/rai-template.yaml` is only asserted to exist, never loaded, even though its own header warns that a misspelled key is dropped in silence.

PR #133, opened 2026-09-10
Recommend test-driven development in guide

renato-umeton:
Added recommendation for test-driven development.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -1,5 +1,7 @@
 # Development Guide
 
+We recommend using test-driven development as much as possible.
+
 ## Setup
 
 This project uses [uv](https://docs.astral.sh/uv/) for environment and dependency management.


DEVELOPMENT.md
```diff
@@ -1,5 +1,7 @@
 # Development Guide
 
+We recommend using test-driven development as much as possible.
```
rafiattrach, line 3: nit: `CONTRIBUTING.md` is probably the better home for this. It already has a `## Testing guidelines` section whose first bullet is "Add tests for any new functionality or bug fixes", which is the same advice one step further along, and `CONTRIBUTING.md` is what a first-time contributor opens first.

Here the sentence sits above `## Setup`, which is the section about installing uv, so it reads as a preamble to the whole document rather than as guidance about testing.

Worth a second thought either way: the repo already asks for tests on every change and CI enforces it, so a line saying to write them first only earns its place if it says something the existing bullet does not. Something like "write the failing test first, then the handler" would at least be concrete.

rafiattrach (changes requested):
Thanks @renato-umeton no argument with the advice. One note on where it lands

DEVELOPMENT.md
renato-umeton: Agreed on the home. #144 moves it to `CONTRIBUTING.md`. Right now it sits in the workflow steps. I will move it into the Testing guidelines list as the first bullet, with the concrete wording: write the failing test first, then the handler. I will close this one in favour of #144.

renato-umeton:
Closed in favour of #144, which puts the sentence in CONTRIBUTING.md as the first Testing guidelines bullet.

PR #134, opened 2026-09-10
fix: refuse unknown RAI config keys and correct the social impact key in the template

### LINKED ISSUE #131: social_ missing?
rai-example.yaml:75 uses social_impact, but rai/loader.py:43 reads data_social_impact, so that field is silently dropped, and --rai-config --help points users at that file. The bundled template uses the correct key..

renato-umeton:
Closes #131.

`rai-example.yaml`, which `--rai-config --help` points users at, used `social_impact` where the loader reads `data_social_impact`, so the field was dropped without a word. This fixes the template and makes the loader refuse the class of mistake instead of hiding it.

## Changes

- `rai-example.yaml` uses `data_social_impact`, and the template now shows every field the schema declares (a source dataset `id` and a platform `url` were missing), guarded by a test that walks every dataclass in `rai/schema.py` against the raw YAML of the shipped template. `models` stays `[]` with a commented example, because an entry asserts that a named model used the dataset, so `ModelRef` is the one stated exemption.
- `rai/loader.py` refuses an unknown key at every level (top level, `ai_fairness`, `lineage`, each source dataset and model, each activity, agent and platform) with a `ValueError` naming the file, the key path (for example `activities[0].agents[1].nme`) and the accepted keys, which are derived from the dataclass fields so they cannot drift. A value that is not the mapping or list its level expects, a YAML syntax error, or a non-mapping document are reported the same way. `collection_types` must be a list of scalars; a bare string used to become a list of characters.
- `collection_types` was accepted and then never written: the same failure one level down. It is now emitted as `rai:dataCollectionType` on the dataset node, the deduplicated union across activities in declaration order. Values are written in the wording RAI 1.0 recommends where one exists (`existing_datasets` becomes `Secondary Data analysis`, `observations` becomes `Passive Data Collection`), the three with no RAI term keep their own name in title case, and any other text is written as given because the range is open text. The template and the guide list the pairs. Only there: RAI 1.0 declares the property on the dataset, not on an activity. The RAI reference output gains exactly that key.
- The CLI reads the config before scanning, including under `--dry-run`, so a bad config fails immediately with exit 1 and no traceback. An intended `typer.Exit` raised inside the generate command's `try` was being swallowed by the catch-all and followed by a bare `Unexpected error:` line; it is re-raised now.
- `docs/user-guide/rai.md` says unknown keys are rejected and that `--dry-run` checks the config too.

The message a user now sees for the issue's typo:

```
Error: rai.yaml: unknown key in the RAI config: ai_fairness.social_impact. Accepted keys at ai_fairness: data_biases, data_limitations, data_social_impact, data_use_cases, has_synthetic_data, personal_sensitive_information. Did you mean 'data_social_impact'?
```

## Release note

Two things a config written against an older template will notice:

- The template's old `social_impact` key is now `data_social_impact`. A config still carrying the old key stops the run. The error names the key and suggests the new one.
- A source dataset or model without `url`, or an agent or platform without `name`, stops the run too. Before, those entries were dropped without a word. An activity without `id` or `type` also stops the run. Before, it was written out with an empty `@id` and a blank label.
- A quoted boolean such as `has_synthetic_data: "false"`, and a list or mapping in a text field, are refused. Before, they were written as `true` or as a Python list repr.
- `collection_types` values are translated on the way out (`observations` is written as `Passive Data Collection`), so a document baked from the same config changes those strings.

## Testing

Test-driven: each behaviour has a test that failed before its commit (`tests/test_rai_loader.py`, additions to `tests/test_rai.py`). `uv run pytest -q`: 1013 passed. `uv run pre-commit run --all-files`: all hooks pass. `docs/generate.py`: no diff. `croissant-baker validate` accepts the updated reference output.

### DIFF
--- a/rai-example.yaml
+++ b/rai-example.yaml
@@ -100,11 +100,13 @@ lineage:
 
   # rai:usedBy — models or assets that have used this dataset (e.g. for training
   # or evaluation). url is required; name and id are optional.
-  # Write `models: []` if no model has used the dataset yet.
-  models:
-    - url: https://huggingface.co/my-org/my-model
-      id: MODEL-001
-      name: My Clinical NLP Model
+  # An entry here asserts that a named model used this dataset, so the shape is
+  # shown as a comment: fill it in only for a model you know of.
+  models: []
+  # models:
+  #   - url: https://huggingface.co/my-org/my-model
+  #     name: My Clinical NLP Model
+  #     id: MODEL-001
 
 
 # ── Activities ────────────────────────────────────────────────────────────────
--- a/rai-example.yaml
+++ b/rai-example.yaml
@@ -72,7 +72,7 @@ ai_fairness:
   # Articulate what new capabilities the dataset enables, who benefits, potential
   # harms, affected groups, and concrete mitigation steps (licensing, access
   # controls, usage agreements).
-  social_impact: >
+  data_social_impact: >
     This dataset enables research that could improve clinical AI tools and
     patient outcomes. However, models trained on biased data risk perpetuating
     health disparities if deployed without careful evaluation. Access is
--- a/rai-example.yaml
+++ b/rai-example.yaml
@@ -90,19 +90,21 @@ lineage:
 
   # prov:wasDerivedFrom — datasets this dataset was derived from.
   # url is required; all other fields are optional.
+  # id is your own identifier for the source, if you use one.
   source_datasets:
     - url: https://physionet.org/content/mimiciii/
+      id: SOURCE-001
       name: MIMIC-III
       organisation: PhysioNet
       license: PhysioNet Credentialed Health Data License 1.5.0
 
   # rai:usedBy — models or assets that have used this dataset (e.g. for training
   # or evaluation). url is required; name and id are optional.
-  models: []
-  # models:
-  #   - url: https://huggingface.co/my-org/my-model
-  #     name: My Clinical NLP Model
-  #     id: MODEL-001
+  # Write `models: []` if no model has used the dataset yet.
+  models:
+    - url: https://huggingface.co/my-org/my-model
+      id: MODEL-001
+      name: My Clinical NLP Model
 
 
 # ── Activities ────────────────────────────────────────────────────────────────
@@ -145,6 +147,7 @@ activities:
         is_synthetic: false
     platforms:
       - name: Hospital Electronic Health Record System
+        url: https://www.bidmc.org
         description: >
           Institutional EHR system used for routine clinical documentation,
           from which the raw data was extracted.
--- a/src/croissant_baker/rai/injector.py
+++ b/src/croissant_baker/rai/injector.py
@@ -36,8 +36,9 @@ def inject_rai(metadata: dict, config: RAIConfig) -> dict:
     - Models that used this dataset → rai:usedBy.
     - Activities → prov:wasGeneratedBy (list of prov:Activity), each with
       optional prov:wasAssociatedWith (agents) and rai:usedPlatform (platforms).
-    - Collection types → rai:dataCollectionType, on each activity that declares
-      one and, unioned, on the dataset node the RAI spec puts the property on.
+    - Collection types → rai:dataCollectionType on the dataset node, unioned
+      across the activities. RAI 1.0 declares the property on sc:Dataset, so
+      it does not go on the prov:Activity that carries the types in the config.
     """
     _ensure_prov_context(metadata, config)
 
@@ -56,8 +57,8 @@ def inject_rai(metadata: dict, config: RAIConfig) -> dict:
     if af.has_synthetic_data is not None:
         metadata["rai:hasSyntheticData"] = af.has_synthetic_data
 
-    # rai:dataCollectionType is a dataset-level property, so the activities'
-    # types are unioned onto the dataset as well as kept on their own nodes.
+    # rai:dataCollectionType is declared on sc:Dataset, so the types every
+    # activity declares are unioned onto the dataset rather than left on it.
     collection_types = _unique(
         t for act in config.activities for t in act.collection_types
     )
@@ -127,8 +128,6 @@ def _build_activity(act: Activity) -> dict:
         node["prov:startedAtTime"] = act.start_at
     if act.end_at:
         node["prov:endedAtTime"] = act.end_at
-    if act.collection_types:
-        node["rai:dataCollectionType"] = _one_or_many(_unique(act.collection_types))
 
     if act.agents:
         agent_nodes = []
--- a/src/croissant_baker/rai/injector.py
+++ b/src/croissant_baker/rai/injector.py
@@ -13,6 +13,16 @@ _ACTIVITY_LABELS = {
 }
 
 
+def _one_or_many(values: list):
+    """Croissant writes a single-valued property as a scalar, not a list."""
+    return values[0] if len(values) == 1 else values
+
+
+def _unique(values) -> list:
+    """The values in declaration order, first occurrence kept."""
+    return list(dict.fromkeys(values))
+
+
 def inject_rai(metadata: dict, config: RAIConfig) -> dict:
     """
     Inject RAI and PROV-O attributes into a Croissant metadata dict.
@@ -26,6 +36,8 @@ def inject_rai(metadata: dict, config: RAIConfig) -> dict:
     - Models that used this dataset → rai:usedBy.
     - Activities → prov:wasGeneratedBy (list of prov:Activity), each with
       optional prov:wasAssociatedWith (agents) and rai:usedPlatform (platforms).
+    - Collection types → rai:dataCollectionType, on each activity that declares
+      one and, unioned, on the dataset node the RAI spec puts the property on.
     """
     _ensure_prov_context(metadata, config)
 
@@ -43,6 +55,14 @@ def inject_rai(metadata: dict, config: RAIConfig) -> dict:
         metadata["rai:dataSocialImpact"] = af.data_social_impact
     if af.has_synthetic_data is not None:
         metadata["rai:hasSyntheticData"] = af.has_synthetic_data
+
+    # rai:dataCollectionType is a dataset-level property, so the activities'
+    # types are unioned onto the dataset as well as kept on their own nodes.
+    collection_types = _unique(
+        t for act in config.activities for t in act.collection_types
+    )
+    if collection_types:
+        metadata["rai:dataCollectionType"] = _one_or_many(collection_types)
 
     # Lineage — source datasets
     if config.lineage.source_datasets:
@@ -68,9 +88,7 @@ def inject_rai(metadata: dict, config: RAIConfig) -> dict:
     # Activities
     activities = [_build_activity(act) for act in config.activities]
     if activities:
-        metadata["prov:wasGeneratedBy"] = (
-            activities[0] if len(activities) == 1 else activities
-        )
+        metadata["prov:wasGeneratedBy"] = _one_or_many(activities)
 
     return metadata
 
@@ -109,6 +127,8 @@ def _build_activity(act: Activity) -> dict:
         node["prov:startedAtTime"] = act.start_at
     if act.end_at:
         node["prov:endedAtTime"] = act.end_at
+    if act.collection_types:
+        node["rai:dataCollectionType"] = _one_or_many(_unique(act.collection_types))
 
     if act.agents:
         agent_nodes = []
@@ -120,9 +140,7 @@ def _build_activity(act: Activity) -> dict:
             if a.description:
                 agent["prov:description"] = a.description
             agent_nodes.append(agent)
-        node["prov:wasAssociatedWith"] = (
-            agent_nodes[0] if len(agent_nodes) == 1 else agent_nodes
-        )
+        node["prov:wasAssociatedWith"] = _one_or_many(agent_nodes)
 
     if act.platforms:
         platform_nodes = []
@@ -133,9 +151,7 @@ def _build_activity(act: Activity) -> dict:
             if p.description:
                 plat["prov:description"] = p.description
             platform_nodes.append(plat)
-        node["rai:usedPlatform"] = (
-            platform_nodes[0] if len(platform_nodes) == 1 else platform_nodes
-        )
+        node["rai:usedPlatform"] = _one_or_many(platform_nodes)
 
     return node
 
--- a/src/croissant_baker/rai/loader.py
+++ b/src/croissant_baker/rai/loader.py
@@ -80,6 +80,31 @@ def _entries(value, path: str, file: Path) -> list[dict]:
     return [_mapping(entry, f"{path}[{i}]", file) for i, entry in enumerate(value)]
 
 
+def _scalars(value, path: str, file: Path) -> list[str]:
+    """Return ``value`` as a list of plain values, dropping the blank ones.
+
+    A bare string is refused rather than iterated: its characters are never
+    what the author meant, and a mapping's keys are not either.
+    """
+    if value is None:
+        return []
+    if not isinstance(value, list):
+        raise ValueError(
+            f"{file}: expected a list of values at {path}, "
+            f"found {type(value).__name__}."
+        )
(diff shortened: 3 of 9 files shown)


rai-example.yaml
```diff
@@ -72,7 +72,7 @@ ai_fairness:
   # Articulate what new capabilities the dataset enables, who benefits, potential
   # harms, affected groups, and concrete mitigation steps (licensing, access
   # controls, usage agreements).
-  social_impact: >
+  data_social_impact: >
```
rafiattrach: This is a breaking change shipping under a `fix:` title, which release-please will turn into a patch.

Before, an unknown key was ignored and the run succeeded. Now it stops the run. The config that breaks is the one the tool itself ships and points users at, so a user who copied `rai-example.yaml` from v0.6.0 and edited the prose goes from a working bake to exit 1 on upgrade:

```
v0.6.0:  social_impact: ...   ->  exit 0, property silently missing
this PR: social_impact: ...   ->  exit 1
```

The error is the better behaviour. The problem is only that a patch release is where nobody looks for it.

Worth either `feat!:` on the squash title or a `BREAKING CHANGE:` footer naming `social_impact` to `data_social_impact`, so it lands in the changelog. The changelog entry matters more than usual here, because upgrading does not repair files that were already published with the property missing, and those users have no other way to find out.

src/croissant_baker/rai/loader.py
```diff
@@ -19,20 +20,107 @@
...
+        return {}
+    if not isinstance(value, dict):
+        raise ValueError(
+            f"{file}: expected a mapping at {path or 'the top level'}, "
+            f"found {type(value).__name__}."
+        )
+    return value
+
+
+def _check_keys(mapping: dict, allowed: frozenset[str], path: str, file: Path) -> None:
```
rafiattrach: Key names are strict now, but values are not checked at all, and one of the coercions publishes the opposite of what the user wrote:

```
has_synthetic_data: false      ->  "rai:hasSyntheticData": false
has_synthetic_data: "false"    ->  "rai:hasSyntheticData": true
has_synthetic_data: "no"       ->  "rai:hasSyntheticData": true
```

That is `bool("false")` being `True`. Quoting the value, which plenty of people do out of habit, inverts a Responsible AI claim, with exit 0 and no message.

A nearby one, same cause:

```
data_biases: [sampling, labelling]   ->  "rai:dataBiases": "['sampling', 'labelling']"
```

A Python list repr in published metadata.

Both behave the same on main, so neither is something this branch introduced. Raising them here because this is the PR that makes RAI config validation strict, and a `_bool` helper that accepts only real YAML booleans plus a type check in `_str` is a few lines next to what you have already written.

nit: the error messages are good but stop just short. At the activity level the user gets eight alphabetised keys to scan:

```
unknown key in the RAI config: activities[0].start_date.
Accepted keys at activities[0]: agents, collection_types, description, end_at, id, platforms, start_at, type.
```

One `difflib.get_close_matches(key, allowed, n=1)` turns that into "did you mean `start_at`?"

src/croissant_baker/rai/loader.py
```diff
@@ -47,59 +135,83 @@ def load_rai_config(path: Path) -> RAIConfig:
...
+    ln_raw = _mapping(raw.get("lineage"), "lineage", path)
+    _check_keys(ln_raw, _LINEAGE_KEYS, "lineage", path)
+
+    source_datasets = []
+    for i, s in enumerate(
+        _entries(ln_raw.get("source_datasets"), "lineage.source_datasets", path)
+    ):
+        _check_keys(s, _SOURCE_DATASET_KEYS, f"lineage.source_datasets[{i}]", path)
+        if not s.get("url"):
+            continue
```
rafiattrach: These `continue` statements drop a whole entry without a word, which is the same failure this PR exists to remove, one level down.

Every key here is spelled correctly:

```yaml
lineage:
  source_datasets:
    - name: MIMIC-III        # no url
      organisation: PhysioNet
activities:
  - id: ACT-001
    type: data_collection
    agents:
      - url: https://example.org/team   # no name
```

Result: exit 0, no message, and the only RAI key in the output is `prov:wasGeneratedBy`. Both entries vanished.

This also makes the sentence at `docs/user-guide/rai.md:121` untrue, and that sentence is new in this PR: "A misspelled key is therefore never dropped in silence, so a field cannot go missing from the output without a word." It can, it just takes a missing required field rather than a misspelled one.

Either raise when a required field is missing, the same way an unknown key now raises, or narrow the claim in the docs to misspelled keys only. The first is better, since the whole point of the PR is that silent loss in RAI metadata is not acceptable.

src/croissant_baker/rai/injector.py
```diff
@@ -44,6 +57,14 @@ def inject_rai(metadata: dict, config: RAIConfig) -> dict:
     if af.has_synthetic_data is not None:
         metadata["rai:hasSyntheticData"] = af.has_synthetic_data
 
+    # rai:dataCollectionType is declared on sc:Dataset, so the types every
+    # activity declares are unioned onto the dataset rather than left on it.
+    collection_types = _unique(
+        t for act in config.activities for t in act.collection_types
+    )
+    if collection_types:
+        metadata["rai:dataCollectionType"] = _one_or_many(collection_types)
```
rafiattrach: This emits a property nothing emitted before. On main, `collection_types` was parsed and then never written out, so this line adds `rai:dataCollectionType` to the output of every existing user who has a `collection_types` block:

```
main:     rai:dataCollectionType absent
this PR:  "rai:dataCollectionType": "observations"
```

The placement reasoning is right, `croissant_rai.ttl` does declare it with `schema:domainIncludes schema:Dataset`. It is just a new feature rather than a fix, which feeds back into the release labelling above.

The part worth changing: the values do not match the vocabulary. RAI 1.0 recommends `Surveys, Secondary Data analysis, Physical data collection, Direct measurement, Document analysis, Manual Human Curator, Software Collection, Experiments, Web Scraping, Web API, Focus groups, Self-reporting, Customer feedback data, User-generated content data, Passive Data Collection, Others`. The template documents `surveys | interviews | observations | experiments | web_scraping | crowdsourcing | existing_datasets | simulations | other`.

While nothing was emitted this did not matter. Now the tool publishes a private snake_case vocabulary where a recommended one exists, and the updated reference fixture pins it. `existing_datasets` is `Secondary Data analysis`, `web_scraping` is `Web Scraping`, and so on. Since the template comment is already being touched in this PR, this is the cheapest moment to align it.

src/croissant_baker/rai/loader.py
rafiattrach, line 130: Key names are strict now, but values are not checked at all, and one of the coercions publishes the opposite of what the user wrote:

```
has_synthetic_data: false      ->  "rai:hasSyntheticData": false
has_synthetic_data: "false"    ->  "rai:hasSyntheticData": true
has_synthetic_data: "no"       ->  "rai:hasSyntheticData": true
```

That is `bool("false")` being `True`. Quoting the value, which plenty of people do out of habit, inverts a Responsible AI claim, with exit 0 and no message.

A nearby one, same cause:

```
data_biases: [sampling, labelling]   ->  "rai:dataBiases": "['sampling', 'labelling']"
```

A Python list repr in published metadata.

Both behave the same on main, so neither is something this branch introduced. Raising them here because this is the PR that makes RAI config validation strict, and a `_bool` helper that accepts only real YAML booleans plus a type check in `_str` is a few lines next to what you have already written.

nit: the error messages are good but stop just short. At the activity level the user gets eight alphabetised keys to scan:

```
unknown key in the RAI config: activities[0].start_date.
Accepted keys at activities[0]: agents, collection_types, description, end_at, id, platforms, start_at, type.
```

One `difflib.get_close_matches(key, allowed, n=1)` turns that into "did you mean `start_at`?"

src/croissant_baker/rai/loader.py
rafiattrach: These `continue` statements drop a whole entry without a word, which is the same failure this PR exists to remove, one level down.

Every key here is spelled correctly:

```yaml
lineage:
  source_datasets:
    - name: MIMIC-III        # no url
      organisation: PhysioNet
activities:
  - id: ACT-001
    type: data_collection
    agents:
      - url: https://example.org/team   # no name
```

Result: exit 0, no message, and the only RAI key in the output is `prov:wasGeneratedBy`. Both entries vanished.

This also makes the sentence at `docs/user-guide/rai.md:121` untrue, and that sentence is new in this PR: "A misspelled key is therefore never dropped in silence, so a field cannot go missing from the output without a word." It can, it just takes a missing required field rather than a misspelled one.

Either raise when a required field is missing, the same way an unknown key now raises, or narrow the claim in the docs to misspelled keys only. The first is better, since the whole point of the PR is that silent loss in RAI metadata is not acceptable.

src/croissant_baker/rai/injector.py
rafiattrach, line 92: This emits a property nothing emitted before. On main, `collection_types` was parsed and then never written out, so this line adds `rai:dataCollectionType` to the output of every existing user who has a `collection_types` block:

```
main:     rai:dataCollectionType absent
this PR:  "rai:dataCollectionType": "observations"
```

The placement reasoning is right, `croissant_rai.ttl` does declare it with `schema:domainIncludes schema:Dataset`.

The part worth changing is the values, which do not match the vocabulary. RAI 1.0 recommends `Surveys, Secondary Data analysis, Physical data collection, Direct measurement, Document analysis, Manual Human Curator, Software Collection, Experiments, Web Scraping, Web API, Focus groups, Self-reporting, Customer feedback data, User-generated content data, Passive Data Collection, Others`. The template documents `surveys | interviews | observations | experiments | web_scraping | crowdsourcing | existing_datasets | simulations | other`.

While nothing was emitted this did not matter. Now the tool publishes a private snake_case vocabulary where a recommended one exists, and the updated reference fixture pins it. `existing_datasets` is `Secondary Data analysis`, `web_scraping` is `Web Scraping`, and so on. Since the template comment is already being touched here, this is the cheapest moment to align it.

rafiattrach (changes requested):
Ran this against the RAI 1.0 vocabulary, mlcroissant's own constants, and a set of configs built to break it.

The central fix is right. `rai:dataSocialImpact` is the canonical name in `croissant_rai.ttl` and in mlcroissant's constants, and the template was reaching the loader under a key it does not read, so the property came out missing. Good catch, and `test_shipped_example_names_every_schema_field` is the right guard, since the existing tests exercised a different fixture that had it spelled correctly.

The loader rewrite reads well too: four small helpers, the accepted keys derived from `schema.py` so there is one source of truth, and errors that name the key by its path in the YAML.

Three things below. One note that does not need a code change: a config carrying the old key now stops the run rather than being ignored, so it is worth a line in the release notes for anyone who copied the template.

src/croissant_baker/rai/loader.py
renato-umeton: Confirmed, on this branch and on main. I will add a `_bool` helper that accepts only real YAML booleans. It covers `has_synthetic_data` and also `is_synthetic` at [loader.py:193](https://github.com/MIT-LCP/croissant-baker/blob/a70474b425467e79f1c124eb2f806ae2a68a93b0/src/croissant_baker/rai/loader.py#L193), which has the same `bool()` call. There a quoted `"false"` turns a human team into `prov:SoftwareAgent`.

For `_str` I will refuse mappings and lists, the same check `_scalars` already makes. I will not require a Python string. YAML reads `start_at: 2011-01-01` as a date and `id: 42` as an int, and both work today. The bare `str()` calls on lines 150, 165, 190, 206, 218 and 219 get the same check, since `url: {href: ...}` leaks a Python repr there too.

The `difflib` hint goes in. It maps `social_impact` to `data_social_impact`, the typo this PR started from.

Separately: `rai:hasSyntheticData` is not in RAI 1.0. It is missing from [croissant_rai.ttl](https://github.com/mlcommons/croissant/blob/main/docs/croissant_rai.ttl) and from [mlcroissant's constants](https://github.com/mlcommons/croissant/blob/main/python/mlcroissant/mlcroissant/_src/core/constants.py), so mlcroissant accepts any value there. The type check is the only guard. Same for `rai:usedBy` and `rai:usedPlatform`. I will open an issue in this repo for that, since the property is ours.

src/croissant_baker/rai/loader.py
renato-umeton: Confirmed. Both of your configs lose the entry with exit 0. Platforms behave the same way at [loader.py:203](https://github.com/MIT-LCP/croissant-baker/blob/a70474b425467e79f1c124eb2f806ae2a68a93b0/src/croissant_baker/rai/loader.py#L203).

I will raise, in the same shape as the unknown-key error: file, YAML path, and the field that is missing. The reason comes from PROV-O. `prov:wasDerivedFrom` and `prov:wasAssociatedWith` need something to identify what they point at, and the loader already treats `url` on a source dataset or model, and `name` on an agent or platform, as that field. The error and the docs will say that.

This reverses what commit 9cfca65 and one test docstring promised, that values stay lenient. No test pins the drop, so nothing else changes. I will also raise for an activity with neither `id` nor `type`, which today is emitted with an empty `@id`.

src/croissant_baker/rai/injector.py
renato-umeton: Agreed on the placement and on the new property. On the values I want to settle the mapping here first, because four of the nine template values have no clean RAI 1.0 term.

Five map directly: `surveys` to Surveys, `experiments` to Experiments, `web_scraping` to Web Scraping, `existing_datasets` to Secondary Data analysis, `other` to Others.

Four do not: `observations`, `interviews`, `crowdsourcing`, `simulations`. The fixture uses `observations`. The nearest terms are Direct measurement or Passive Data Collection for observations, Focus groups or Surveys for interviews, Manual Human Curator for crowdsourcing, and Software Collection for simulations. Each of those changes the meaning.

The list in the [RAI spec](https://docs.mlcommons.org/croissant/docs/croissant-rai-spec.html) is a recommendation that ends in Others, and mlcroissant accepts free text there. So my plan: change the template and the docs to the RAI wording, map the five clean values, and pass anything else through as written. The reference fixture and the tests that assert the snake_case values change with it. If you would rather have a fixed mapping for all nine, let me know which terms to use for the four.

renato-umeton:
I replied in each of the three threads. On the release note: the changelog comes from the squash commit, so I will put the line about `social_impact` becoming `data_social_impact` in the PR description and next to the unknown-keys section in `docs/user-guide/rai.md`. The new errors for a missing `url` or `name` go in the same two places. I will not use a `BREAKING CHANGE` footer for either. The release config does not set `bump-minor-pre-major`, so that footer would publish 1.0.0 instead of 0.6.1 ([release-please docs](https://github.com/googleapis/release-please/blob/main/docs/manifest-releaser.md)).

The branch still merges cleanly onto main after 0.6.0. I merged it locally and the suite passes on the result.

src/croissant_baker/rai/injector.py
rafiattrach: Five map straight onto the RAI recommended values:

  surveys           -> Surveys
  experiments       -> Experiments
  web_scraping      -> Web Scraping
  existing_datasets -> Secondary Data analysis
  other             -> Others

observations -> Passive Data Collection fits as well.

For the last three there is no RAI equivalent. Since the property says
"Recommended values" and its range is schema:Text it isn't a closed list,
so I'd rather keep the honest term than force a bad fit:

  interviews    -> Interviews
  crowdsourcing -> Crowdsourcing
  simulations   -> Simulations

Mapping interviews onto Focus groups would lose the distinction between a
one-to-one interview and a group session, and there's no benefit to the
consumer in doing it.

what do you think?

renato-umeton: Thx, went with exactly that. The nine values map on the way out, case and spacing do not matter (`SURVEYS` and `Web_Scraping` still land on `Web Scraping` and `Surveys`), and anything else is written as given since the range is open text. Template and guide show the pairs, and a test holds the injector table, the template block and the docs table together so they cannot drift. The reference fixture now carries `Passive Data Collection` and `Secondary Data analysis`. Pushed, 1013 tests green.

PR #135, opened 2026-09-10
ci: publish test coverage as a README badge and pull request comment

### LINKED ISSUE #132: add modern test coverage visibility
README.md could use a coverage % pill, with a link to test results

renato-umeton:
Closes #132.

Adds a coverage badge to the README, linked to a browsable HTML report, produced entirely inside GitHub Actions with the built-in token. No external service or secret is needed.

## How it works

- `pytest-cov` joins the `test` group, with `[tool.coverage.run]` (`source = ["croissant_baker"]`, branch coverage, `relative_files`) and `[tool.coverage.report]` in `pyproject.toml`. Locally: `uv run pytest --cov --cov-report=term-missing`. Measured today: 92% (3930 statements, 1420 branches).
- `test.yaml` runs pytest with `--cov` on both matrix legs and, on the 3.12 leg only, calls `py-cov-action/python-coverage-comment-action@v4.3`. On a push to `main` the action writes `endpoint.json` and `htmlcov/` to the `python-coverage-comment-action-data` branch; on a pull request it posts a coverage comment. `persist-credentials: false` on the checkout and job-scoped `permissions` follow the action's v4 setup.
- `coverage-comment.yaml` is the action's documented second workflow, triggered by `workflow_run`, so pull requests from forks get the comment too (the first workflow only has a read token there and saves the comment as an artifact). Concurrency is grouped by `head_branch`, since `github.ref` is always `main` on `workflow_run`.
- The test workflow is grouped by workflow and ref without `cancel-in-progress`, so two quick pushes to `main` queue rather than race the data-branch commit.
- The README badge is a shields.io endpoint on that branch, linking to the HTML report through htmlpreview. It renders as an error until the first push to `main` creates the branch, which happens at merge time; `DEVELOPMENT.md` says so and documents the coverage command.

## Testing

Test-first: `tests/test_coverage_setup.py` (14 tests) pins the cross-file invariants that would otherwise rot silently: the coverage configuration in `pyproject.toml`, the `pytest-cov` floor (`>=6`, since `exclude_also` needs coverage 7.2+), the `--cov` step and the action step in `test.yaml`, the single matrix leg that publishes, the artifact name and path handshake between the two workflows (a rename in one place would silently end fork-PR comments, because the action exits 0 on a missing artifact), the `workflow_run` trigger naming the test workflow exactly, and the README badge and link both naming the same data branch and repository. The tests run on both matrix legs (`tomli` fallback on 3.10).

`uv run pytest -q`: 965 passed. `uv run pre-commit run --all-files`: all hooks pass. `uv lock --check` passes; the lock diff adds only `coverage` and `pytest-cov` plus the stale project version.

Nothing for maintainers to configure. The action is pinned by tag to match the repository's other action pins; its README now recommends SHA pins, which would be a repository-wide change.

### DIFF
--- a/.github/workflows/coverage-comment.yaml
+++ b/.github/workflows/coverage-comment.yaml
@@ -11,21 +11,29 @@ on:
     types:
       - completed
 
+concurrency:
+  # On workflow_run, github.ref is always the default branch, so group by the
+  # pull request's branch instead. Otherwise one pull request cancels another.
+  group: ${{ github.workflow }}-${{ github.event.workflow_run.head_branch }}
+  cancel-in-progress: true
+
+# The job below states what it needs.
+permissions: {}
+
 jobs:
   post-comment:
     name: Post coverage comment
     runs-on: ubuntu-latest
     if: github.event.workflow_run.event == 'pull_request' && github.event.workflow_run.conclusion == 'success'
     permissions:
-      # Publish the comment.
+      # Post the comment, and edit it on later runs.
       pull-requests: write
-      # Edit the existing comment instead of adding a second one.
-      contents: write
-      # Look up the run that triggered this one and download its artifact.
+      # Download the comment artifact from the run that triggered this one.
       actions: read
+      contents: read
     steps:
       - name: Post comment
-        uses: py-cov-action/python-coverage-comment-action@v3.41
+        uses: py-cov-action/python-coverage-comment-action@v4.3
         with:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           GITHUB_PR_RUN_ID: ${{ github.event.workflow_run.id }}
--- a/.github/workflows/coverage-comment.yaml
+++ b/.github/workflows/coverage-comment.yaml
@@ -1,0 +1,31 @@
+name: Post coverage comment
+
+# A pull request from a fork runs with a read-only token, so "Run Tests" can
+# only save the comment as an artifact. This workflow runs from the base
+# repository once that one finishes and posts it. It deliberately never checks
+# the contributor's code out:
+# https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
+on:
+  workflow_run:
+    workflows: ["Run Tests"]
+    types:
+      - completed
+
+jobs:
+  post-comment:
+    name: Post coverage comment
+    runs-on: ubuntu-latest
+    if: github.event.workflow_run.event == 'pull_request' && github.event.workflow_run.conclusion == 'success'
+    permissions:
+      # Publish the comment.
+      pull-requests: write
+      # Edit the existing comment instead of adding a second one.
+      contents: write
+      # Look up the run that triggered this one and download its artifact.
+      actions: read
+    steps:
+      - name: Post comment
+        uses: py-cov-action/python-coverage-comment-action@v3.41
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_PR_RUN_ID: ${{ github.event.workflow_run.id }}
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -9,6 +9,13 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
+    permissions:
+      # Publish or update the coverage comment on a trusted pull request.
+      # GitHub downgrades both of these to read-only for runs from forks, which
+      # is why coverage-comment.yaml exists.
+      pull-requests: write
+      # Push the coverage data branch that feeds the README badge.
+      contents: write
     # Without this the default is six hours, and a suite that hangs burns all
     # of it on every matrix job before anyone sees a log.
     timeout-minutes: 20
@@ -29,4 +36,22 @@ jobs:
         run: uv sync --group test
 
       - name: Run tests
-        run: uv run pytest -v
+        run: uv run pytest -v --cov --cov-report=xml --cov-report=term-missing:skip-covered
+
+      # The badge and the comment describe one run, so only the 3.12 leg of the
+      # matrix reports; the 3.10 leg still measures coverage, it just drops it.
+      - name: Coverage comment
+        id: coverage_comment
+        if: matrix.python-version == '3.12'
+        uses: py-cov-action/python-coverage-comment-action@v3.41
+        with:
+          GITHUB_TOKEN: ${{ github.token }}
+
+      - name: Store pull request comment to be posted
+        if: steps.coverage_comment.outputs.COMMENT_FILE_WRITTEN == 'true'
+        uses: actions/upload-artifact@v7
+        with:
+          # Renaming either of these means setting COMMENT_ARTIFACT_NAME or
+          # COMMENT_FILENAME on the action in both workflows.
+          name: python-coverage-comment-action
+          path: python-coverage-comment-action.txt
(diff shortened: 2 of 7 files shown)


pyproject.toml
```diff
@@ -65,5 +67,19 @@ pythonpath = ["src"]
...
+[tool.coverage.run]
+source = ["croissant_baker"]
+branch = true
+# python-coverage-comment-action reads the .coverage file from a checkout whose
+# paths differ from the runner's, so record repository-relative paths.
+relative_files = true
+
+[tool.coverage.report]
+show_missing = true
+exclude_also = [
```
rafiattrach, line 79: Both of these patterns are already in coverage's built-in defaults, so this block does nothing. `exclude_also` appends to `DEFAULT_EXCLUDE`, and on the version this repo locks (7.16.0) that default already is:

```
#\s*(pragma|PRAGMA)[:\s]?\s*(no|NO)\s*(cover|COVER)
^\s*(((async )?def .*?)?[\])]+(\s*->.*?)?:\s*)?\.\.\.\s*(#|$)
if (typing\.)?TYPE_CHECKING:
```

I ran `coverage report` with the block and with it deleted:

```
with:     TOTAL  3930  243  1420  169  92%
without:  TOTAL  3930  243  1420  169  92%
```

Identical. `pragma: no cover` has been a default since coverage 4.x, and `if TYPE_CHECKING:` since 7.10.0.

The knock-on cost is the reason to act on it rather than leave it. The comment raises the `pytest-cov` floor to `>=6` specifically so `exclude_also` works, and `tests/test_coverage_setup.py` then adds a test to defend that floor. Dropping the block removes all three.

README.md
```diff
@@ -6,6 +6,7 @@
 
 <p align="center">
   <a href="https://github.com/MIT-LCP/croissant-baker/actions/workflows/test.yaml"><img src="https://github.com/MIT-LCP/croissant-baker/actions/workflows/test.yaml/badge.svg" alt="CI"></a>
+  <a href="https://htmlpreview.github.io/?https://github.com/MIT-LCP/croissant-baker/blob/python-coverage-comment-action-data/htmlcov/index.html"><img src="https://img.shields.io/endpoint?url=https://raw.githubusercontent.com/MIT-LCP/croissant-baker/python-coverage-comment-action-data/endpoint.json" alt="Coverage"></a>
```
rafiattrach, line 9: This badge will render orange from its first run, which reads as a warning on the front page of a healthy repo.

The action defaults `MINIMUM_GREEN` to 100 and `MINIMUM_ORANGE` to 70 (both confirmed in `action.yml` on the v4.3 tag), and neither workflow sets either one. Coverage measures 92%, so it lands in the orange band.

`MINIMUM_GREEN: 90` on the action step in both workflows would fix it. Whatever number we pick, it is worth setting deliberately rather than inheriting 100 at least for now

.github/workflows/coverage-comment.yaml
```diff
@@ -0,0 +1,39 @@
...
+on:
+  workflow_run:
+    workflows: ["Run Tests"]
+    types:
+      - completed
+
+concurrency:
+  # On workflow_run, github.ref is always the default branch, so group by the
+  # pull request's branch instead. Otherwise one pull request cancels another.
+  group: ${{ github.workflow }}-${{ github.event.workflow_run.head_branch }}
```
rafiattrach, line 17: `head_branch` is the branch name inside the contributor's fork, and fork branch names collide constantly. With `cancel-in-progress: true`, two open pull requests whose fork branches happen to share a name land in the same group and the newer run cancels the older one, so a contributor loses their coverage comment for no reason.

This is not hypothetical here. #133 in this same repo is on a branch called `patch-1`, which is the name GitHub assigns to every edit made through the web UI. Any two of those collide.

Adding the fork's identity separates them:

```yaml
group: ${{ github.workflow }}-${{ github.event.workflow_run.head_repository.full_name }}-${{ github.event.workflow_run.head_branch }}
```

.github/workflows/test.yaml
```diff
@@ -6,9 +6,25 @@ on:
...
+# Queue rather than race: two pushes to main in quick succession would otherwise
+# have both runs committing to the coverage data branch at once. Nothing is
+# cancelled, so a run in progress always finishes and comments.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+
 jobs:
   test:
     runs-on: ubuntu-latest
+    permissions:
```
rafiattrach, line 21: These write scopes are granted to the whole matrix job, so they are live during `Run tests`, which executes the pull request's own test code and runs `uv sync` build hooks. They are also granted to the 3.10 leg, which never reaches the coverage step at all.

Exposure is limited in practice, since `GITHUB_TOKEN` is not exported to `run:` steps and `persist-credentials: false` keeps it out of `.git/config`, so this is hardening rather than an open hole. The tidier shape is a second job with `needs: test` that downloads the `.coverage` artifact and runs only the action, leaving the test job at `permissions: {}`.

Separately, and worth one line in `DEVELOPMENT.md`: nothing gates on coverage. There is no `fail_under` and no threshold on the action, so coverage can fall from 92% to 20% with every check still green. That is a reasonable choice for a first coverage PR, but the docs describe the badge and the delta comment without saying the number is advisory, and a reader will assume CI enforces it.

tests/test_coverage_setup.py
```diff
@@ -0,0 +1,156 @@
...
+    # Two legs would upload the same artifact name and race each other's commit
+    # to the data branch.
+    assert "matrix.python-version ==" in _coverage_step(_workflow("test.yaml"))["if"]
+
+
+def test_the_test_workflow_queues_runs_instead_of_cancelling_them() -> None:
+    # A cancelled run is a lost coverage comment, and two runs pushing the data
+    # branch at once is a lost commit, so runs queue and none is cancelled.
+    concurrency = _workflow("test.yaml")["concurrency"]
+    assert concurrency["group"] == "${{ github.workflow }}-${{ github.ref }}"
```
rafiattrach, line 155: nit: about half this file tests real invariants and half asserts that a config file says what it says.

Worth keeping, because no single edit keeps them consistent: the `"checkout" not in yaml.dump(comment)` check (a genuine security invariant), the `workflow_run.workflows` list matching `test.yaml`'s `name`, the artifact name matching the action's default, and the badge URL matching `COVERAGE_DATA_BRANCH`.

The ones that cannot catch a regression: asserting `source == ["croissant_baker"]` and `relative_files is True` just reads the TOML back and compares it to the TOML. Nobody changes those by accident, and anyone who changes them deliberately edits this file in the same commit. A test that ran coverage and asserted the `.coverage` database holds relative `src/croissant_baker/*` paths and no test files would actually be checking something.

Two brittle spots while you are in here:

```python
assert concurrency["group"] == "${{ github.workflow }}-${{ github.ref }}"   # exact string
floor = int(spec.removeprefix("pytest-cov>=").split(".")[0])              # ValueError on "pytest-cov>=6,<8"
```

The first turns red on any harmless refinement of the group. The second raises rather than failing with a message the moment someone adds an upper bound.

rafiattrach (changes requested):
Ran the coverage setup locally and read both workflows against the action's own source and docs.

The security side is right, which is the part this pattern usually gets wrong. No `pull_request_target` anywhere, the `workflow_run` workflow checks out nothing and runs no shell step, `permissions: {}` at the top with narrow job scopes, and `persist-credentials: false` on the checkout, which is a subtle one that the action's v4.1 notes actually require. Coverage comes out at 92% and the `.coverage` file carries relative paths, so what the workflows consume matches what `pyproject.toml` configures.

Four things to change before merge, none of them large. The first two are visible to anyone who lands on the README.

Issue #136, opened 2026-09-10
add support for chemistry and chemoinformatics files


renato-umeton:
with genomics these would be also very useful, and the README should list those file formats too

PR #139, opened 2026-09-10
feat: add SMILES and XYZ handlers for chemistry data

renato-umeton:
Part of #136. The structure and small-molecule formats named there (PDB, mmCIF, MOL, SDF) are handled by #141 through gemmi; this PR covers the two formats that do not overlap with it.

Adds handlers for SMILES and XYZ, with no new dependencies. Each reads a header, a first frame, or a stated bounded sample and stops; the read is bounded by the format, never by the file size. The README's supported-formats table lists both. Stacked on #127 (the genomic handlers): the diff against `main` includes that branch until it merges, and the first commits unique to this PR start after `620c0d5`.

## SMILES

`handlers/smiles_handler.py` (`.smi`, `.smiles`) claims a file whose first molecule line carries a plausible SMILES token, or a header line followed by one (RDKit's `SmilesWriter` and the ChEMBL, ZINC and Enamine exports all write `SMILES Name` first). From a bounded sample (1000 lines or 1 MiB) it reports the delimiter, the column count (capped at 300, stated) and the header names when present, and emits one RecordSet of `sc:Text` fields. Nothing from the data lines is emitted. Media type `chemical/x-daylight-smiles`.

## XYZ

`handlers/xyz_handler.py` (`.xyz`) reads the first frame's count and comment line and checks the first atom line: atom count, comment, extended-XYZ property names when the comment declares `Properties=`. Frames are not counted and no coordinate is read. A zero-atom frame is legal. Media type `chemical/x-xyz`.

Fields of the SMILES RecordSet carry `source` and no `extract`, as the VCF handler's do, because mlcroissant cannot read the format and a column reference would be a promise nothing can keep.

## Supporting changes

- `handlers/utils.py` gains `PrefixLines`, one bounded chunk-to-lines reader now used by SMILES, XYZ and the SAM handler, replacing the hand-rolled loops. A trailing partial line at end of stream is delivered as a line. The reader tells a stream that ends exactly on its byte bound from one the bound cut short, at the cost of one probe byte, so a file that fits the bound exactly is described rather than refused.
- Digit runs are guarded before `int()` in XYZ, so a refusal always names the file rather than surfacing CPython's integer-conversion limit.
- The legacy-contract test double formerly named for `.xyz` now uses an extension no built-in owns.
- Guide sections, README rows and the generated formats table for both formats.

## History

The branch first added MOL, SDF, PDB and mmCIF handlers as well. They were removed once #141 was on the table, since two handlers per extension in the registry is a coin toss over which one describes a file. The commits that added them are still in the history, so a squash merge is preferable to a merge commit: release-please would otherwise list features that never shipped.

## Testing

- One test module per handler (claims on suffix plus content, refusals with reasons, CRLF, no trailing newline, bounded reads asserted with a counting stream, field caps, wrapped inputs, refusals driven through a bake and read from the scan report, full bakes constructing under mlcroissant), plus `SAMPLES` entries driving the contract sweep and the compression matrix.
- Validated against files written by RDKit (SMILES with and without a header) and ASE (plain, extended and multi-frame XYZ); every extracted value matched the writing tool, and CRLF conversions changed nothing.
- `uv run pytest -q`: 1451 passed. `uv run pre-commit run --all-files`: all hooks pass. `docs/generate.py`: no diff.


renato-umeton:
Done. This PR now targets feat/genomic-handlers, which I pushed to MIT-LCP at #127's head (620c0d5), so the diff shows only the chemistry commits. When #127 merges I will rebase this onto main and point it back at main.

renato-umeton:
Rebased onto the new #127 head, so this stays a stack of the chemistry commits only. Suite green at 1583.

PR #141, opened 2026-09-11
feat: structural biology files (PDB, mmCIF, STAR, MRC, MTZ, mdoc, SDF)

### LINKED ISSUE #138: add support for structural biology files
I'm thinking of adding libraries like Gemmi (fast, handles mmCIF/PDB/MTZ/CCP4 maps cleanly, and is the best single tool for format conversion), Biotite and Biopython for parsing, MDAnalysis and MDTraj for trajectories, mrcfile for volumes, starfile and mdocfile for cryo-EM metadata, OpenBabel/RDKit for small molecules...

renato-umeton:
# Structural biology files

Closes #138.

## What this changes

Six handlers, in one package, `handlers/structural_biology/`:

| Format | Extensions | What is described |
| - | - | - |
| Macromolecular structure | `.pdb`, `.ent`, `.cif`, `.mmcif` | Entry id, title, experimental method, resolution, unit cell, space group, and model, chain, residue and atom counts. A small-molecule CIF by its cell, space group and formula. A CIF holding no structure (a dictionary, a deposition log) as the tables it declares. |
| STAR | `.star` | One record set per data block, RELION optics and particles included, with columns typed from their values. |
| MRC / CCP4 map | `.mrc`, `.mrcs`, `.map`, `.ccp4` | Grid dimensions, data type, voxel size, space group, and whether the file is a volume, an image stack or a volume stack. |
| MTZ | `.mtz` | One typed field per reflection column, unit cell, space group, resolution range and datasets. |
| SerialEM mdoc | `.mdoc` | Acquisition globals and the per-section keys of a tilt series or montage. |
| Small molecules | `.sdf`, `.mol`, `.mol2` | Molecule, atom and bond counts, and the property tags an SDF library declares, typed. |

One optional dependency, `gemmi`, for PDB, mmCIF and STAR parsing, installed as the `structural-biology` extra: `pip install "croissant-baker[structural-biology]"`. It ships wheels for every platform and Python version this package supports. Without it the CLI still runs; a PDB, mmCIF or STAR file is refused with one line saying what to install, and every other format here is described as before. MRC and MTZ read a fixed header with the standard library and never touch the data; mdoc, SDF, MOL and MOL2 are read in one forward pass with no library at all.

No field carries an `extract`. mlcroissant has no reader for any of these formats, so every field states its source file and, for tables, no value is ever emitted, following the GEO SOFT and HDF5 handlers.

Every handler reads through the decompressed stream, so `model.cif.gz` is described exactly as `model.cif` is. The demo dataset includes one such file.

## Design notes

**One `.cif` suffix, three dictionaries.** An mmCIF entry writes dotted category tags, a small-molecule CIF writes the underscore-only tags of the core dictionary, and a dictionary or log carries no atoms at all. The tags decide. Structures share one FileSet and one `structures` record set because they answer the same questions; a document that holds no structure gets its per-table record sets instead, and the FileSet excludes it by path so the per-file record set counts structures only.

**Only the header for MRC and MTZ.** A 5 GB map costs 1024 bytes to describe. The MTZ header is found through the offset in its second word and parsed to `END`.

**Trajectories are out of scope.** XTC, TRR and DCD would need MDAnalysis or MDTraj, whose current releases target Python 3.11 or later while this package supports 3.10, and a trajectory is the one case in this family where the frames rather than the header are what a consumer wants. The user guide says so.

## Tests

Test-first throughout. Every handler has its own test module covering header values against files written by gemmi and by hand, both MRC endiannesses, the `MAP ` and `MTZ ` signatures, garbage bytes, truncated headers, upper-case suffixes, missing files, empty batches and id disambiguation across directories. The contract sweep and the compression matrix cover all six through `SAMPLES`.

A demo dataset under `tests/data/input/structural_biology_demo/` is written by a committed deterministic builder, baked into a golden, validated with mlcroissant, and compared in both discovery orders. It prints `Scanned 15 file(s): 14 described, 1 not described`, the one being its README.

## Size

| | Lines |
| - | - |
| `src/` | +2,264 |
| tests | +4,049 / -1 |
| fixtures and golden | +1,693 |
| docs and README | +274 / -1 |

42 files. The suite goes from 951 to 1,391 tests; without gemmi installed, 83 of them skip and none error.


renato-umeton:
Rebased on main at 0.6.0. The test helpers and the end-to-end test file conflicted and both sides were kept. The lock is relocked at main's revision with the structural-biology extra, and the golden gained the @id main now emits. Suite green at 1523.

PR #140, opened 2026-09-11
feat: digital pathology whole-slide image formats (SVS, NDPI, SCN, BIF, QPTIFF, DICOM WSI)

renato-umeton:
Closes #137.

## Summary

Adds header-only support for the digital pathology whole-slide image formats that cover the bulk of public datasets (TCGA, CPTAC, CAMELYON and most vendor exports), with no new dependency.

**Vendor TIFF slides** (new `WSIHandler`, backed by the pure reader `handlers/wsi.py`):

| Extension | Scanner |
|-----------|---------|
| `.svs` | Aperio (Leica Biosystems) |
| `.ndpi` | Hamamatsu NanoZoomer |
| `.scn` | Leica SCN400 |
| `.bif` | Roche / Ventana iScan |
| `.qptiff` | Akoya / PerkinElmer Vectra |

Each slide is read through the already-required `tifffile` and described by vendor, base dimensions, pyramid level count, tile size, microns per pixel, objective magnification, compression and the associated images it carries (label, macro, thumbnail). The reader mirrors the OME-XML policy: DTD or entity declarations, oversized descriptions and malformed vendor XML are refused with a named reason and the file is still described from its TIFF tags.

**DICOM whole-slide microscopy** (extends the existing `DICOMHandler`): instances of the VL Whole Slide Microscopy Image SOP class additionally record image flavor (VOLUME, LABEL, OVERVIEW, THUMBNAIL), total pixel matrix size, imaged volume size, slide container identifier, and pixel spacing from the shared functional groups. Output for non-slide DICOM is byte-identical to before; all existing goldens are unchanged.

## Not in this PR

MIRAX (`.mrxs`), Olympus VSI, Zeiss CZI, Philips iSyntax, OME-Zarr and DZI are documented as not yet supported. The first two and the last two are directories rather than files and need a directory-as-one-unit discovery concept the baker does not have yet; CZI and iSyntax need a new library. Annotation formats are a separate problem and are out of scope.

## Testing

Test-driven throughout: each behaviour was written as a failing test first. New tests cover the reader per vendor, the handler contract sweep (claims, magic, compression wrappers, case), record-set shape, DICOM slide fields, and an end-to-end bake of a committed synthetic dataset (`tests/data/input/wsi_demo`, one zero-pixel slide per vendor plus two DICOM slide instances, under 70 KB total) validated with mlcroissant and compared against a golden across both discovery orders.

## Docs

[...]

PR #142, opened 2026-09-11
docs: group the README formats table by field

renato-umeton:
## Summary

Add a leading **Field** column to the supported formats table in `README.md` and group the rows by it: General, Clinical records, Physiological signals, Radiology, Pathology, Genomics, Chemistry and chemoinformatics, Structural biology. Row order within a field is unchanged, and no format, extension or note text changes.

## Depends on all other open PRs by @renato-umeton

The grouped table covers every handler in #127, #128, #130, #133, #134, #135, #139, #140 and #141, so this branch is built on top of all of them: it is upstream `main` with those nine PRs merged in PR-number order, followed by the README change. Merge those first; this PR then reduces to the README table (plus two small merge fixups listed below).

Conflict resolutions made while stacking the PRs:

- `uv.lock` regenerated with `uv lock` after #130 and #135.
- #140 and #141 both extend `README.md`, `tests/helpers.py`, `tests/test_end_to_end.py`, `tests/test_utils.py` and `handlers/utils.py`; both sides are kept.
- `handlers/registry.py`: the chemistry handlers from #139 are listed before the structural biology handlers from #141.
- `docs/_generated/formats-table.md` regenerated with `docs/generate.py`.
- Two blank lines restored in `tests/helpers.py` so `ruff format` passes.

## Known test failures

#139 and #141 both register handlers for `.pdb`, `.ent`, `.cif`, `.mmcif`, `.sdf` and `.mol`, and registration order decides ownership. With the chemistry handlers first, 8 tests fail (2360 pass): the structural biology end-to-end golden, the handler contract routing tests for those extensions, and the two routing tests in #141. Ordering the other way fails 15. Reconciling those two PRs is what fixes this; it is outside the scope of the README change.

## Test plan

- `uv run pytest`: 2360 passed, 8 failed as described above.
- `uv run ruff check .` and `uv run ruff format --check .`: clean.
- `uv lock --check`: passes.


renato-umeton:
Converting to draft. This groups the README table by field and needs the rows from #127, #139, #140 and #141, so it carries all of them until they land. I will rebuild it on main once they merge.

